PR #1, opened 2017-11-30
[WIP] Support for Prophecy

Jean85:
Reading https://github.com/phpstan/phpstan/issues/134 and wanting to try PHPStan 0.9, I wanted to try to write the extension to make PHPStan and Prophecy work along.

I'm not an expert on PHPStan internals, so I started working on this copypasting from the MockBuilder classes, and I reached a point where the `reveal()` method is no longer an issue for PHPStan! :tada: 

Now, I need to tackle the issues reported when defining the expectation on a mock, and that's a lot trickier!

Scenario:

```php
$mock = $this->prophesize(SUTClass::class);
$mock->methodA(...)
    ->should...()
    ->will...();

```
Basically you can call on the mock any method of it's base class (`ObjectProphecy`) but also any method of the mocked class (`methodA`); the problem is that those methods will accept the same stuff defined bu the original signature OR `TokenInterface`s from Prophecy, OR a mix of that.; also, it will return a `MethodProphecy` every time.

How we should tackle this?

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -18,3 +18,11 @@ services:
 		class: PHPStan\Type\PHPUnit\MockBuilderDynamicReturnTypeExtension
 		tags:
 			- phpstan.broker.dynamicMethodReturnTypeExtension
+	-
+		class: PHPStan\Type\Prophecy\ProphetDynamicReturnTypeExtension
+		tags:
+			- phpstan.broker.dynamicMethodReturnTypeExtension
+	-
+		class: PHPStan\Type\Prophecy\ObjectProphecyDynamicReturnTypeExtension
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
Hi, there's already an extension for that 😊 https://github.com/Jan0707/phpstan-prophecy

Discussed in depth here: https://github.com/phpstan/phpstan/issues/586

Feel free to check out if it does the job better or worse than yours and try to join forces with the author 😊

Jean85:
Damn, I missed that!
Yep, it seems to work pretty good! I'll dive into the code ASAP, thanks!

PR #2, opened 2017-12-10
Add missing methods that generate mocks

lcobucci:
Some methods were missing, and when used `phpstan` doesn't work properly. It shows errors like:

```sh
Property MyTest::$property (MockedClass&PHPUnit\Framework\MockObject\MockObject) does not accept PHPUnit\Framework\MockObject\MockObject. 
```

This PR fixes it, but I didn't add any test (since the repo doesn't have any)


ondrejmirtes:
Thank you! 👍

PR #6, opened 2017-12-27
AssertEqualsIsDiscouraged

mhujer:
This rule is even more opinionated than the previous ones.

If you don't think it should be included in PHPStan, I can change it and keep only the first part (which suggests `assertSame` in for the same types)

### DIFF
--- a/src/Rules/PHPUnit/AssertEqualsIsDiscouragedRule.php
+++ b/src/Rules/PHPUnit/AssertEqualsIsDiscouragedRule.php
@@ -1,0 +1,94 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\PHPUnit;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Type\BooleanType;
+use PHPStan\Type\FloatType;
+use PHPStan\Type\IntegerType;
+use PHPStan\Type\StringType;
+use PHPStan\Type\VerbosityLevel;
+
+/**
+ * @implements \PHPStan\Rules\Rule<\PhpParser\NodeAbstract>
+ */
+class AssertEqualsIsDiscouragedRule implements \PHPStan\Rules\Rule
+{
+
+	public function getNodeType(): string
+	{
+		return \PhpParser\NodeAbstract::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		if (!AssertRuleHelper::isMethodOrStaticCallOnAssert($node, $scope)) {
+			return [];
+		}
+
+		/** @var \PhpParser\Node\Expr\MethodCall|\PhpParser\Node\Expr\StaticCall $node */
+		$node = $node;
+
+		if (count($node->args) < 2) {
+			return [];
+		}
+		if (!$node->name instanceof Node\Identifier || strtolower($node->name->name) !== 'assertequals') {
+			return [];
+		}
+
+		$leftType = $scope->getType($node->args[0]->value);
+		$rightType = $scope->getType($node->args[1]->value);
+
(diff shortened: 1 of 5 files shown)


src/Rules/PHPUnit/AssertEqualsIsDiscouragedRule.php
```diff
@@ -0,0 +1,79 @@
...
+			return [];
+		}
+
+		$leftType = $scope->getType($node->args[0]->value);
+		$rightType = $scope->getType($node->args[1]->value);
+
+		if (
+			($leftType instanceof BooleanType && $rightType instanceof BooleanType)
+			|| ($leftType instanceof IntegerType && $rightType instanceof IntegerType)
+			|| ($leftType instanceof StringType && $rightType instanceof StringType)
```
ondrejmirtes: You might be able to simplify this condition with `Type::isSupersetOf()` and a UnionType consisting of bool, int and string. cc @JanTvrdik might help you with that.

src/Rules/PHPUnit/AssertEqualsIsDiscouragedRule.php
```diff
@@ -0,0 +1,79 @@
...
+				sprintf(
+					'You should use assertSame instead of assertEquals, because both values are of the same type "%s"',
+					$typeDescription
+				),
+			];
+		}
+		if (
+			($leftType instanceof FloatType && $rightType instanceof FloatType)
+			&&
+			count($node->args) < 4 // is not using delta for comparing floats
```
ondrejmirtes: This newline isn't necessary.

src/Rules/PHPUnit/AssertEqualsIsDiscouragedRule.php
```diff
@@ -0,0 +1,79 @@
...
+		) {
+			return [
+				'You should use assertSame instead of assertEquals, because both values are of the same type "float" and you are not using $delta argument',
+			];
+		}
+
+		$fileContents = explode("\n", file_get_contents($scope->getFile()));
+		$previousLine = $fileContents[$node->getLine() - 2];
```
ondrejmirtes: I don't like this. You should look into `$node` properties instead, there should be a comment available.

ondrejmirtes (changes requested):

ondrejmirtes:
I submitted a review 😊 Also please rebase your branch on master and fix PHP 7.1-related CS errors.

src/Rules/PHPUnit/AssertEqualsIsDiscouragedRule.php
mhujer: I'm not sure if the union would mean the same thing here. Both (left and right) need to be of the same type.

mhujer:
R4R @ondrejmirtes (I kept the fixes in the separate commits for easier review)

mhujer:
Any feedback on this one?

Majkl578:
TBH I don't think this belongs here directly, it should be in phpstan-strict-rules (or i.e. phpstan-phpunit-strict-rules), or at least in separate ruleset (i.e. `strict-rules.neon`).

src/Rules/PHPUnit/AssertEqualsIsDiscouragedRule.php
Jean85: Maybe union + `\getClass($leftType) === \getClass($rightType)`?

ikvasnica:
I just made a very similar rule to my package, check it out and feel free to comment. 

It forbids `assertEquals` with any scalar type, no matter whether the two arguments are of the same type or not. 

https://github.com/ikvasnica/phpstan-clean-test/pull/12

PR #8, opened 2017-12-30
Expand AssertSameWithCountRule to support sizeof

carusogabriel:
As suggested in #7, I've expand `AssertSameWithCountRule` to support `count`'s alias: `sizeof` :+1:

### DIFF
--- a/src/Rules/PHPUnit/AssertSameWithCountRule.php
+++ b/src/Rules/PHPUnit/AssertSameWithCountRule.php
@@ -36,19 +36,14 @@ class AssertSameWithCountRule implements \PHPStan\Rules\Rule
 		if (
 			$right instanceof Node\Expr\FuncCall
 			&& $right->name instanceof Node\Name
+			&& in_array(strtolower($right->name->toString()), ['count', 'sizeof'], true)
 		) {
-			$method = strtolower($right->name->toString());
-			$message = 'You should use assertCount($expectedCount, $variable) instead of assertSame($expectedCount, %s($variable)).';
-
-			if ($method === 'count') {
-				return [
-					sprintf($message, 'count'),
-				];
-			} elseif ($method === 'sizeof') {
-				return [
-					sprintf($message, 'sizeof'),
-				];
-			}
+			return [
+				sprintf(
+					'You should use assertCount($expectedCount, $variable) instead of assertSame($expectedCount, %s($variable)).',
+					strtolower($right->name->toString())
+				),
+			];
 		}
 
 		return [];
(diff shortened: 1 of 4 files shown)


src/Rules/PHPUnit/AssertSameWithCountRule.php
```diff
@@ -36,11 +36,19 @@ public function processNode(Node $node, Scope $scope): array
 		if (
 			$right instanceof Node\Expr\FuncCall
 			&& $right->name instanceof Node\Name
-			&& strtolower($right->name->toString()) === 'count'
 		) {
-			return [
-				'You should use assertCount($expectedCount, $variable) instead of assertSame($expectedCount, count($variable)).',
-			];
+			$method = strtolower($right->name->toString());
```
ondrejmirtes: It's not a `method`, it's a `functionName`.

src/Rules/PHPUnit/AssertSameWithCountRule.php
```diff
@@ -36,11 +36,19 @@ public function processNode(Node $node, Scope $scope): array
...
+			$message = 'You should use assertCount($expectedCount, $variable) instead of assertSame($expectedCount, %s($variable)).';
+
+			if ($method === 'count') {
+				return [
+					sprintf($message, 'count'),
```
ondrejmirtes: What about `sprintf($message, $method)` instead of this if/else?

carusogabriel: Shouldn’t we verify if the function is `count` or `sizeof`? I followed the if that was here 😁

ondrejmirtes: OK, just try to simplify the code somehow :)

carusogabriel: Okay :man_technologist:

src/Rules/PHPUnit/AssertSameWithCountRule.php
```diff
@@ -36,10 +36,13 @@ public function processNode(Node $node, Scope $scope): array
 		if (
 			$right instanceof Node\Expr\FuncCall
 			&& $right->name instanceof Node\Name
-			&& strtolower($right->name->toString()) === 'count'
+			&& in_array(strtolower($right->name->toString()), ['count', 'sizeof'])
```
carusogabriel: @ondrejmirtes Should I assigned the `functionName` inside the if, or is fine to repeat `strtolower($right->name->toString()`?

Issue #11, opened 2018-01-21
Require phpunit on dev deps


soullivaneuh:
PHPUnit should not be a mandatory requirement.

Some users would like to use PHPUnit phar global installation and don't want to have to get the whole dependencies of PHPUnit and itself on the project vendor dir.

What is the goal of this requirement? I don't see any direct class usage so far.

UPDATE: BTW, it also mandatory to require also phpstan from plugins? What is the usage if we use the docker image or the shim repository?

Regards

ondrejmirtes:
Hi, there's a similar request on phpstan-doctrine: https://github.com/phpstan/phpstan-doctrine/pull/9 I'm still evaluating this. The problem is that PHPStan needs to see PHPUnit classes to analyse code correctly. So if you don't have them autoloaded, you need to add your PHPUnit installation to `autoload_directories`/`autoload_files`.

phpstan-shim is fine since it has `phpstan/phpstan` in compoer.json `replace`: https://github.com/phpstan/phpstan-shim/blob/master/composer.json#L8 So when you're installing an extension alongside it, it does not install `phpstan/phpstan` but reuses the shim version.

Docker: I'm not sure about how to solve this one, I'll ask the author of the images.

Soullivaneuh:
> phpstan-shim is fine since it has phpstan/phpstan in compoer.json replace: 

You didn't understand IMHO, I'm talking about global installation like `composer global require phpstan/phpstan-shim`, or a `.phar`.

If we require one of the plugins on our project dependencies, we will have phpstan installed (and phpunit) for this case.

Sample: `symfony/phpunit-bridge` works with PHPUnit but do not directly require it. It let the choice to the user. :+1: 

 >  So if you don't have them autoloaded, you need to add your PHPUnit installation to autoload_directories/autoload_files.

Well, not a big deal if this is configurable. We just have to add the correct error message. ;-)

o5:
👍  for `dev` deps.

I wanted to use PHPUnit 7 in our global composer dependencies (in Docker base image) and it failed:

```
Problem 1
    - Installation request for phpstan/phpstan-phpunit 0.9.3 -> satisfiable by phpstan/phpstan-phpunit[0.9.3].
[...]

PR #13, opened 2018-02-02
Allow PHPUnit 7

o5:
https://github.com/phpstan/phpstan-phpunit/issues/11#issuecomment-362503290

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
 	"require": {
 		"php": "~7.1",
 		"phpstan/phpstan": "^0.10",
-		"phpunit/phpunit": "^6.3"
+		"phpunit/phpunit": "^7.0 || ^6.3"
 	},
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Hi, this PR had number of missing things so I decided to come up with my own approach: https://github.com/phpstan/phpstan-phpunit/commit/852411f841a37aeca2fa5af0002b0272c485c9bf

Specifically:

* The change had to be based on 0.9.3 and not on master so it could be released quickly
* I write the lower version first and upper second: `^6.3 || ^7.0`
* The Travis build had to be updated so all version constraints were tested.

It was released as 0.9.4 just a few moments ago.

o5:
@ondrejmirtes I can confirm, version 0.9.4 resolves my problem. Thank you!

PR #37, opened 2019-01-19
Enhancement: Cache dependencies installed with composer between builds

localheinz:
This PR

* [x] caches dependencies installed with `composer` between builds


ondrejmirtes:
This is obsolete, we already switched to GitHub Actions.

PR #40, opened 2019-01-20
Fix: No need to update composer itself twice

localheinz:
This PR

* [x] stops updating `composer` itself twice on Travis


ondrejmirtes:
This is obsolete, we already switched to GitHub Actions.

PR #47, opened 2019-05-10
Fix detection of unspecific assert*() with direct Assert::assert*() calls

### LINKED ISSUE #46: Unspecific assert methods not detected when called on Assert class directly
The following is detected inside a test and `assertNull` is suggested:
```php
$this->assertSame(null, $foo);
self::assertSame(null, $foo);
```
However the following is not:
```php
Assert::assertSame(null, $foo);
```

Majkl578:
Fixes #46.


ondrejmirtes:
Thanks 😊

ondrejmirtes:
Released as 0.11.1

Majkl578:
@ondrejmirtes Thanks!

PR #50, opened 2019-07-08
add getMockForTrait to dynamic return type

pavarnos:


pavarnos:
This fix is different to #29. It does not reply on the fluent phpUnit mock builder like #29 does. It seems to work fine on my code base.

ondrejmirtes:
Hi, it can't work since trait isn't a typehintable type. You'd have to create an artificial interface with all the trait's methods which probably isn't worth it.

PR #51, opened 2019-07-16
add php 7.3, 7.4snapshot

andreybolonin:


ondrejmirtes:
I'd accept PHP 7.3 if you submit it in a separate PR. PHP 7.4 isn't released yet and the build failed. Please make first sure that the build succeeds and then submit a PR.

PR #59, opened 2020-01-19
Allow extending only from specified classes in unit tests

ikvasnica:
Hi, this experimental PR adds a new strict rule to allow extending only from PHPUnit's TestCase in unit tests. E.g. `class MyTest extends AbstractKernelTestCase` will fail.

**Issue to be fixed:** It should fix the problem when developers use DI containers in unit tests and overuse inheritance by using custom abstract classes.

Please, consider this as a work-in-progress. I will be very glad for any comment or suggestion (it is my first real open source PR). Thanks :)


ondrejmirtes:
Hi, this is too opinionated for phpstan-phpunit. It's not a bug to extend from `AbstractKernelTestCase`. Feel free to contribute this for example to https://github.com/ergebnis/phpstan-rules or to create your own package.

There is nothing wrong about the rule, but it should be more configurable.

ikvasnica:
Thanks for your feedback.

I made the rule more configurable and make my own library from it: https://github.com/ikvasnica/phpstan-clean-test

PR #65, opened 2020-03-20
Assertion class as parameter for Codeception config

### LINKED ISSUE #34: assertInternalType()
I'm using [Codeception's wrapper](https://github.com/Codeception/Codeception/blob/2.5/src/Codeception/Util/Shared/Asserts.php#L448-L451) of [PHPUnit's `assertInternalType`](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Assert.php#L1677), and thought phpstan would understand the type.

Example code:
```php
$feedHandle = fopen('php://temp', 'rw+');
$this->assertInternalType('resource', $feedHandle);
fwrite($feedHandle, $feed);
```

Actual result:
```
Parameter #1 $fp of function fwrite expects resource, resource|false given.
```

Expected result:
-- Nothing --

Slamdunk:
Resolves https://github.com/phpstan/phpstan-phpunit/issues/34

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -20,10 +20,14 @@ services:
 			- phpstan.typeSpecifier.functionTypeSpecifyingExtension
 	-
 		class: PHPStan\Type\PHPUnit\Assert\AssertMethodTypeSpecifyingExtension
+		arguments:
+			classWithAssertionMethods: PHPUnit\Framework\TestCase
 		tags:
 			- phpstan.typeSpecifier.methodTypeSpecifyingExtension
 	-
 		class: PHPStan\Type\PHPUnit\Assert\AssertStaticMethodTypeSpecifyingExtension
+		arguments:
+			classWithAssertionMethods: PHPUnit\Framework\Assert
 		tags:
 			- phpstan.typeSpecifier.staticMethodTypeSpecifyingExtension
 	-
--- a/README.md
+++ b/README.md
@@ -94,3 +94,39 @@ To perform framework-specific checks, include also this file:
 ```
 
 </details>
+
+
+## Codeception parameters
+
+This extension works well with [Codeception](https://github.com/Codeception/Codeception) too.
+
+[Unit tests](https://codeception.com/docs/05-UnitTests) are already recognised by default;
+[Functional tests](https://codeception.com/docs/04-FunctionalTests) and
+[Acceptance tests](https://codeception.com/docs/03-AcceptanceTests) instead have to be configured
+manually since assertion classes are generated at runtime with different namespaces and class names.
+
+Here's an example configuration:
+
+```
+services:
+    -
+        class: PHPStan\Type\PHPUnit\Assert\AssertMethodTypeSpecifyingExtension
+        arguments:
+            classWithAssertionMethods: My\CustomNamespace\_support\FunctionalTester
+        tags:
+            - phpstan.typeSpecifier.methodTypeSpecifyingExtension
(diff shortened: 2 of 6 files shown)


extension.neon
```diff
@@ -20,10 +20,14 @@ services:
 			- phpstan.typeSpecifier.functionTypeSpecifyingExtension
 	-
 		class: PHPStan\Type\PHPUnit\Assert\AssertMethodTypeSpecifyingExtension
+		arguments:
+			classWithAssertionMethods: PHPUnit\Framework\TestCase
```
Slamdunk: I've retained the class used in https://github.com/phpstan/phpstan-phpunit/pull/65/files#diff-afc1ebcbfd95543266d01a244f0e43deL27, but shouldn't be this `PHPUnit\Framework\Assert` too?

ondrejmirtes:
Hi, I don't know anything about Codeception. Do you typically use `assert*` methods from PHPUnit in conjunction with Codeception methods, or do you use either PHPUnit or Codeception?

Anyway, we could register everything by default (there's no downside) and just mention in the README that this extension works well with Codeception out of the box. WDYT?

Slamdunk:
It was my wish from the beginning, but it not feasible.

Codeception ships with 2 different types of test, the Unit one which is a subclass of `PHPUnit\Framework\TestCase` and already works out of the box, and the Functional/Acceptance ones. In a nutshell, Functional/Acceptance tests gather multiple classes and their methods, copy-paste their signatures into a giant class, and proxy all the calls to the original ones. This allows the user to choose which module to use in the tests (i.e. Filesystem, Redis, Symfony, FTP...) and call all the available assertions/helper-methods under the `$this` hat. The Assertion module is one of them and is a proxy to the PHPUnit one.

What's the blocking? The giant class created lies under the project folder tree, and of course to be correctly autoloaded it must have the project's namespace, not a generic one.

nreynis:
I feel like this is related to https://github.com/phpstan/phpstan-phpunit/issues/103  
  
I don't exactly see the benefit to parametrize the assertion class instead of just fixing the FQCN. But I'm not familiar at all with codeception. Can you provide a link to that `Acceptance` class (or other relevant code) is it extending `Assert`?

Slamdunk:
> Can you provide a link to that `Acceptance` class (or other relevant code) is it extending `Assert`?

I doesn't exist, Codeception doesn't work this way.
It generates Traits that proxy the calls to PHPUnit's `Assert` class, and it generates them dynamically based on how the developer configured Codeception.

There is no way `phpstan/phpstan-phpunit` can fix #34 with an hardcoded code change.

nreynis:
The thing is making the class parametrizable will fix your use case by breaking PHP Unit analysis.

I think a better approach would be to create another package dedicated to codeception, dependant on this one. This way you can have two instance of the extension working at the same time.

Slamdunk:
> The thing is making the class parametrizable will fix your use case by breaking PHP Unit analysis.

May I ask you how this break would happen?

> I think a better approach would be to create another package dedicated to codeception, dependant on this one. This way you can have two instance of the extension working at the same time.

This PR already enable to have both PHPUnit and Codeception working in the same time, where do you see this isn't the case?

nreynis:
🤦 You are right I didn't understand the PR correctly. I somehow thought you were just feeding the existing service with another default value. But you are in fact creating other instances with different parameters.  
  
This should work fine. 👍

Issue #68, opened 2020-04-09
Improve check for assertSame() instead of assertCount().


VincentLanglet:
Actually we got an error for 
```
assertSame(1, count($variable))
```

Would it be possible to also get an error for
```
assertSame(1, $variable->count())
```
if $variable is `Countable` ?

ondrejmirtes:
Yes, it would be really easy to implement it. This is the current rule: https://github.com/phpstan/phpstan-phpunit/blob/master/src/Rules/PHPUnit/AssertSameWithCountRule.php

VincentLanglet:
@ondrejmirtes Indeed, it was not so hard, the PR is opened https://github.com/phpstan/phpstan-phpunit/pull/69 :)

PR #69, opened 2020-04-16
Add rule for countable

VincentLanglet:
Close https://github.com/phpstan/phpstan-phpunit/issues/68

### DIFF
--- a/src/Rules/PHPUnit/AssertSameWithCountRule.php
+++ b/src/Rules/PHPUnit/AssertSameWithCountRule.php
@@ -4,6 +4,7 @@ namespace PHPStan\Rules\PHPUnit;
 
 use PhpParser\Node;
 use PHPStan\Analyser\Scope;
+use PHPStan\Type\ObjectType;
 
 /**
  * @implements \PHPStan\Rules\Rule<\PhpParser\NodeAbstract>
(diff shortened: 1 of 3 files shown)


src/Rules/PHPUnit/AssertSameWithCountRule.php
```diff
@@ -44,6 +45,21 @@ public function processNode(Node $node, Scope $scope): array
...
 
+		if (
+			$right instanceof Node\Expr\MethodCall
+			&& $right->name instanceof Node\Identifier
+			&& strtolower($right->name->toString()) === 'count'
+			&& count($right->args) === 0
+		) {
+			$type = $scope->getType($right->var);
+
+			if ($type instanceof ObjectType && $type->isInstanceOf(\Countable::class)->yes()) {
```
ondrejmirtes, line 56: There should be a different condition: `(new ObjectType(\Countable::class))->isSuperTypeOf($type)->yes()`. It accounts for some special cases, like `StaticType`, and I like to be consistent.

VincentLanglet: @ondrejmirtes Done :)

ondrejmirtes:
Released as 0.12.8 :)

PR #72, opened 2020-05-07
Report error when trying to configure a non existing method on MockObject

VincentLanglet:


ondrejmirtes:
Just pushed some commits here.

VincentLanglet:
Thanks @ondrejmirtes. The PR is RTM now.

I don't like the `getClassName` checks, `0` offset and `array_filter` manipulation I had to done.
If you have a better way, I take it !

VincentLanglet:
Test are fine now. :)

src/Type/PHPUnit/MockObjectDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,42 @@
...
+	public function getTypeFromMethodCall(MethodReflection $methodReflection, MethodCall $methodCall, Scope $scope): Type
+	{
+		$type = $scope->getType($methodCall->var);
+		if (!($type instanceof IntersectionType)) {
+			return new GenericObjectType(InvocationMocker::class, []);
+		}
+
+		$mockClass = array_filter($type->getTypes(), function (Type $type): bool {
+			return !($type instanceof ObjectType && $type->getClassName() === MockObject::class);
+		});
```
ondrejmirtes, line 37: Make sure the array isn't empty; please name the variable as a plural.

ondrejmirtes: Can you change the condition into:

```php
!$type instanceof TypeWithClassName || $type->getClassName() !== MockObject::class
```

ondrejmirtes:
Oops - one more thing - the new rule isn't registered in rules.neon.

src/Type/PHPUnit/MockObjectDynamicReturnTypeExtension.php
VincentLanglet: > Make sure the array isn't empty

why ?

I'll return 
```
return new GenericObjectType(InvocationMocker::class, []);
```

This sounds ok to me

ondrejmirtes: That's invalid.

VincentLanglet:
This should be ok now @ondrejmirtes :)

ondrejmirtes:
Thank you!

src/Rules/PHPUnit/MockMethodCallRule.php
```diff
@@ -0,0 +1,85 @@
...
+
+namespace PHPStan\Rules\PHPUnit;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\Generic\GenericObjectType;
+use PHPStan\Type\IntersectionType;
+use PHPStan\Type\ObjectType;
+use PHPUnit\Framework\MockObject\InvocationMocker;
```
Tobion, line 11: This class does not exist and might be causing https://github.com/phpstan/phpstan-phpunit/issues/73

VincentLanglet: Should be fixed in https://github.com/phpstan/phpstan-phpunit/pull/74

Tests was running on phpunit ^7.0. The class was existing.

PR #74, opened 2020-05-30
Add Phpunit 9 compatibility

VincentLanglet:


ondrejmirtes:
How is it that PHPUnit 7 has `PHPUnit\Framework\MockObject\InvocationMocker` but all of the versions also have `PHPUnit\Framework\MockObject\Builder\InvocationMocker`? Isn't it a different class with different purpose?

ondrejmirtes:
I can see the stub is already for the builder one, interesting...

PR #79, opened 2020-07-18
Add ShouldCallParentMethodsRule

### LINKED ISSUE #78: [Feature Request] New rule for making sure parent setUp or tearDown methods are called if extending custom base test case
It's common to create a base TestCase that defines some logic in `setUp` and `tearDown` methods of PHPUnit. Laravel, Symfony etc. does this.

So when a user is overriding the `setUp` or `tearDown` in a specific test class, forgetting to call `parent::setUp()` can cause some issues. (I got bitten by this today. Forgot to call `parent::tearDown` so database transactions were not committed. Apparently Laravel does this in it's `tearDown`)

It'd be good to have a rule that catches that. It should be a little smart about it, and only warn when parent `setUp` or `tearDown` is not the base PHPUnit TestCase.

If this is something that you see value in, I can try to implement it.

canvural:
Closes #78

### DIFF
--- a/rules.neon
+++ b/rules.neon
@@ -3,3 +3,4 @@ rules:
 	- PHPStan\Rules\PHPUnit\AssertSameNullExpectedRule
 	- PHPStan\Rules\PHPUnit\AssertSameWithCountRule
 	- PHPStan\Rules\PHPUnit\MockMethodCallRule
+	- PHPStan\Rules\PHPUnit\ShouldCallParentMethodsRule
--- a/rules.neon
+++ b/rules.neon
@@ -1,3 +1,11 @@
+conditionalTags:
+    PHPStan\Rules\PHPUnit\ShouldCallParentMethodsRule:
+        phpstan.rules.rule: %featureToggles.bleedingEdge%
+
+services:
+    -
+        class: PHPStan\Rules\PHPUnit\ShouldCallParentMethodsRule
+
 rules:
 	- PHPStan\Rules\PHPUnit\AssertSameBooleanExpectedRule
 	- PHPStan\Rules\PHPUnit\AssertSameNullExpectedRule
--- a/rules.neon
+++ b/rules.neon
@@ -11,4 +11,3 @@ rules:
 	- PHPStan\Rules\PHPUnit\AssertSameNullExpectedRule
 	- PHPStan\Rules\PHPUnit\AssertSameWithCountRule
 	- PHPStan\Rules\PHPUnit\MockMethodCallRule
-	- PHPStan\Rules\PHPUnit\ShouldCallParentMethodsRule
--- a/src/Rules/PHPUnit/ShouldCallParentMethodsRule.php
+++ b/src/Rules/PHPUnit/ShouldCallParentMethodsRule.php
@@ -21,9 +21,6 @@ class ShouldCallParentMethodsRule implements \PHPStan\Rules\Rule
 
 	public function processNode(Node $node, Scope $scope): array
 	{
-		/** @var InClassMethodNode $node */
-		$node = $node;
-
 		if ($scope->getClassReflection() === null) {
 			return [];
 		}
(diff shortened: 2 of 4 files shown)


rules.neon
```diff
@@ -3,3 +3,4 @@ rules:
 	- PHPStan\Rules\PHPUnit\AssertSameNullExpectedRule
 	- PHPStan\Rules\PHPUnit\AssertSameWithCountRule
 	- PHPStan\Rules\PHPUnit\MockMethodCallRule
+	- PHPStan\Rules\PHPUnit\ShouldCallParentMethodsRule
```
ondrejmirtes, line 6: Instead of adding this as a new rule which would block another 0.12.x release, could you:

1) Require PHPStan 0.12.33 as it contains the bleedingEdge parameter
2) Add this as a conditional rule like this:

```
conditionalTags:
    PHPStan\Rules\PHPUnit\ShouldCallParentMethodsRule:
        phpstan.rules.rule: %featureToggles.bleedingEdge%

services:
    -
        class: PHPStan\Rules\PHPUnit\ShouldCallParentMethodsRule
```

src/Rules/PHPUnit/ShouldCallParentMethodsRule.php
```diff
@@ -0,0 +1,104 @@
...
+
+	public function getNodeType(): string
+	{
+		return InClassMethodNode::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		/** @var InClassMethodNode $node */
+		$node = $node;
```
ondrejmirtes, line 25: This isn't necessary, PHPStan already knows the type.

src/Rules/PHPUnit/ShouldCallParentMethodsRule.php
```diff
@@ -0,0 +1,104 @@
...
+
+			if (strtolower($class) !== 'parent') {
+				continue;
+			}
+
+			if (! $stmt->expr->name instanceof Node\Identifier) {
+				continue;
+			}
+
+			if (in_array(strtolower($stmt->expr->name->name), ['setup', 'teardown'], true)) {
```
ondrejmirtes, line 96: This would accept parent::setUp() in tearDown method, and vice-versa. We need to be more strict.

tests/Rules/PHPUnit/ShouldCallParentMethodsRuleTest.php
```diff
@@ -0,0 +1,32 @@
...
+	protected function getRule(): Rule
+	{
+		return new ShouldCallParentMethodsRule();
+	}
+
+	public function testRule(): void
+	{
+		$this->analyse([__DIR__ . '/data/missing-parent-method-calls.php'], [
+			[
+				'Missing call to parent::setUp method.',
```
ondrejmirtes, line 22: I'd like 'Missing call to parent::setUp() method.'. more

rules.neon
ondrejmirtes: Once we have it in conditionalTags, it should no longer be in the rules section.

tests/Rules/PHPUnit/ShouldCallParentMethodsRuleTest.php
```diff
@@ -0,0 +1,35 @@
...
+		return new ShouldCallParentMethodsRule();
+	}
+
+	public function testRule(): void
+	{
+		$this->analyse([__DIR__ . '/data/missing-parent-method-calls.php'], [
+			[
+				'Missing call to parent::setUp() method.',
+				32,
+			],[
```
ondrejmirtes, line 24: This should be formatted the same way as the next item (open square bracket on the next line)

ondrejmirtes:
Thank you, I'll tag it once I have a computer again :)

ondrejmirtes:
Tagged as 0.12.14 :)

canvural:
Great! Thanks.

ondrejmirtes:
FYI found out very quickly the logic wasn't quite right, just fixed it :) https://github.com/phpstan/phpstan-phpunit/commit/264abf4c789118c63389bb8009e3418f7cd5fbbd

PR #80, opened 2020-07-28
Use the parent Assert class for earlyTerminatingMethodCalls

kbond:
Just getting started with phpstan so forgive me if this won't work as I expect.

I have several custom objects that have phpunit assertions. I call `PHPUnit\Framework\Assert::fail()` from them and was getting false positive return missing errors. Since `PHPUnit\Framework\TestCase` extends from `PHPUnit\Framework\Assert`, I believe this won't break anything.


ondrejmirtes:
Makes sense, thank you :)

PR #91, opened 2021-03-05
Allow toggling union to intersection type conversion

ruudk:
Same as https://github.com/phpstan/phpstan-mockery/pull/28

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -1,4 +1,6 @@
 parameters:
+	phpunit:
+		convertUnionToIntersectionType: true
 	additionalConstructors:
 		- PHPUnit\Framework\TestCase::setUp
 	earlyTerminatingMethodCalls:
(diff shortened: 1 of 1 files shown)


extension.neon
```diff
@@ -1,4 +1,6 @@
 parameters:
+	phpunit:
+		convertUnionToIntersectionType: true
 	additionalConstructors:
 		- PHPUnit\Framework\TestCase::setUp
 	earlyTerminatingMethodCalls:
```
ruudk, line 6: @ondrejmirtes I can create a PR to move these properties to `phpunit` structure if you want.

ondrejmirtes: This is a setting from core PHPStan that this extension adds to.

ondrejmirtes:
Thank you

Issue #87, opened 2020-12-10
Support for methods added with addMethods phpunit function


lugus:
Hello, thank you for your work.

Reporting issue about the `onlyMethods` from `phpunit`, it looks than `phpstan-phpunit` do not check methods added with these method and raise an error.

When:

```
$fooMock = $this->getMockBuilder(\stdClass::class)
                ->addMethods(['hello'])
                ->getMock();
$fooMock->method('hello')
        ->willReturn('Hello PHPStan');
```

```
 ------ -----------------------------------------------------------------------------------------------
  Line   Tests/Unit/FooTest.php
 ------ -----------------------------------------------------------------------------------------------
  120    Trying to mock an undefined method hello() on class stdClass.
 ------ -----------------------------------------------------------------------------------------------
```

VincentLanglet:
I just encountered this issue with the code
```
$object2 = $this->getMockBuilder(\stdClass::class)->addMethods(['addFooBar'])->getMock();
$object2->expects($this->once())->method('addFooBar')->with($instance);
```

Any recommendation @ondrejmirtes ?

VincentLanglet:
I created a reproducer: https://github.com/phpstan/phpstan-phpunit/pull/93

ondrejmirtes:
I don't understand the use-case. If you're adding methods to the mock that don't exist on the mocked class, you'll also get PHPStan errors where you're calling the passed object in production code. e.g. `Method addFooBar does not exist on stdClass.`. Isn't it better to define an interface that requires those methods to be implemented?

VincentLanglet:
The usecase is 
```
$fieldDescription = $this->createMock(FieldDescriptionInterface::class);
$fieldDescription->expects($this->once())->method('getAssociationMapping')->willReturn(['fieldName' => 'fooBar']);
$fieldDescription->expects($this->once())->method('getParentAssociationMappings')->willReturn([]);

$instance = new \stdClass();
$object = $this->getMockBuilder(\stdClass::class)->setMethods(['addFooBar'])->getMock();
$object->expects($this->once())->method('addFooBar')->with($instance);

ObjectManipulator::addInstance($object, $instance, $fieldDescription);
```

Here, the method name `addFooBar` is dynamically generated because of the fieldName `'fooBar'`.
There is no and never will interface for this.

lugosium:
Some situations where i use `addMethods`:

[...]

PR #93, opened 2021-03-22
Avoid addMethod issue with stdClass

VincentLanglet:
Since the tests are written with phpunit@7, I used `setMethods()`.
But the issue also exists with `addMethods()`

### DIFF
--- a/src/Rules/PHPUnit/MockMethodCallRule.php
+++ b/src/Rules/PHPUnit/MockMethodCallRule.php
@@ -52,6 +52,12 @@ class MockMethodCallRule implements \PHPStan\Rules\Rule
 				return $class !== MockObject::class;
 			});
 
+			// `MockBuilder::setMethods` and `MockBuilder::addMethods` are not supported by this rule
+			// so since a lot of the usage are with `stdClass`, we silent the error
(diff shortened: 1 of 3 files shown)


stubs/MockBuilder.stub
```diff
@@ -26,4 +26,14 @@ class MockBuilder
 	 */
 	public function getMockForAbstractClass() {}
 
+	/**
+     * @phpstan-return MockBuilder
+     */
+    public function setMethods(array $methods = null) {}
```
VincentLanglet, line 32: @ondrejmirtes I thought this would have fixed https://github.com/phpstan/phpstan-phpunit/issues/87

I would have expect that the MockBuilder will lose his template param
```
$this->getMockBuilder(Bar::class)
```
would be MockBuilder<Bar> so
```
$this->getMockBuilder(Bar::class)->getMock()
```
would be MockObject&Bar but I wanted that
```
$this->getMockBuilder(Bar::class)->setMethods(...)
```
become something like MockBuilder<mixed> so
```
$this->getMockBuilder(Bar::class)->setMethods(...)->getMock()
```
become MockObject&mixed and every method are allowed

This could be maybe a first step for solving the issue.
But it seems like I'm doing this wrong...

PR #102, opened 2021-08-12
Add support for assertObjectHasAttribute

jeroennoten:
This adds support for `assertObjectHasAttribute()`, so that PHPStan will not complain when writing tests like:

```php
assertObjectHasAttribute('foo', $someObject);
assertSame(123, $someObject->foo);
```


ondrejmirtes:
Thank you!

PR #110, opened 2021-12-17
Ensure the value of `$actual` is not constant

johnbillion:
It's easy to accidentally get the order of the `$expected` and `$actual` parameters reversed in assertions. While the assertion still functions as expected, if the assertion fails then the error message is misleading because the expected and actual results are reversed.

```php
$expected = 123;
$actual = foo();

// Which of these is correct?
self::assertSame( $expected, $actual );
self::assertSame( $actual, $expected);
```

Here's a real world example where the WordPress core test suite was corrected after the discovery of several reversed `$expected` and `$actual` parameters: https://core.trac.wordpress.org/changeset/52248

This PR introduces a rule which triggers an error if the `$actual` parameter is an instance of `ConstantType`. While this doesn't cover every possibility of reversing these parameters, it does cover a potentially common one.


ondrejmirtes:
Hi, this seems too strict for the purpose of phpstan-phpunit. The tests still work, it's just that the message won't be as clear. Pretty minor thing I'd say :)

Also I can't merge it in a minor version without the bleedingEdge flag. Feel free to publish this as your own package :) Thank you.

PR #119, opened 2022-02-19
Validate coverage tags point to real classes

neclimdul:
Its easy to make typos or miss updating a covers tag during a refactor and this helps catch it earlier.

### DIFF
--- a/src/Rules/PHPUnit/CoversShouldExistRule.php
+++ b/src/Rules/PHPUnit/CoversShouldExistRule.php
@@ -1,0 +1,185 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\PHPUnit;
+
+use PhpParser\Comment\Doc;
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\PhpDocParser\Ast\PhpDoc\PhpDocNode;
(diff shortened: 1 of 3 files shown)


ondrejmirtes (changes requested):
Hi! Thank you, this looks really useful.

Could you please make these modifications?

1) Instead of getting PHPDoc AST directly by parsing it like this, ask for it through `PHPStan\Type\FileTypeMapper`. Some details: https://phpstan.org/developing-extensions/reflection#retrieving-custom-phpdocs
2) Run the rule only when bleedingEdge is on. Here's an example how the config should look like: https://github.com/phpstan/phpstan-doctrine/commit/00e7f8f97b2172793006e72da48714e881da6567 Actually right now you don't run the rule at all, so please modify the config according to the linked commit :)
3) Check the PHPDocs only in classes where it's relevant, e.g. child classes of PHPUnit's Testcase.

mad-briller:
@neclimdul is this something you are interested in finishing or should i pick this up?

this was a custom rule we wrote into our previous static analysis tool that we lost since adopting phpstan and would love to see it completed

neclimdul:
@mad-briller still interested but might be a while before I can finish it so if you've got some time to bring it home that would be awesome.

ondrejmirtes:
Superseded by https://github.com/phpstan/phpstan-phpunit/pull/137

Issue #120, opened 2022-03-04
False positive in PHPStan 1.4.8 with `assertNotSame` with instances of the same class


oliverklee:
Using phpstan/phpstan-phpunit 1.0.0 and phpstan/phpstan 1.4.8, the following code creates a false positive:

```php
namespace OliverKlee\Oelib\Tests\Unit\Templating;

use Nimut\TestingFramework\TestCase\UnitTestCase;
use OliverKlee\Oelib\Templating\Template;
use OliverKlee\Oelib\Templating\TemplateRegistry;

class TemplateRegistryTest extends UnitTestCase
{
    /**
     * @test
     */
    public function getForExistingTemplateFileNameCalledTwoTimesReturnsNewInstance(): void
    {
        self::assertNotSame(
            TemplateRegistry::get('EXT:oelib/Tests/Functional/Fixtures/Template.html'),
            TemplateRegistry::get('EXT:oelib/Tests/Functional/Fixtures/Template.html')
        );
    }
}
```

This is the error (with different line numbers):

```
 ------ --------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   Tests/Unit/Templating/TemplateRegistryTest.php                                                                                                           
 ------ --------------------------------------------------------------------------------------------------------------------------------------------------------- 
  69     Call to static method PHPUnit\Framework\Assert::assertNotSame() with OliverKlee\Oelib\Templating\Template and OliverKlee\Oelib\Templating\Template will  
         always evaluate to false.                                                                                                                                
  91     Call to static method PHPUnit\Framework\Assert::assertNotSame() with OliverKlee\Oelib\Templating\Template and OliverKlee\Oelib\Templating\Template will  
         always evaluate to false.                                                                                                                                
 ------ --------------------------------------------------------------------------------------------------------------------------------------------------------- 
```

This problem is new with PHPStan 1.4.8 (as it did no occur with PHPStan 1.4.7).

You can find the complete build with the problem at https://github.com/oliverklee/ext-oelib/pull/904.

(I was not able to use the PHPStan playground for this as I currently do not know of any way of including PHPUnit in it: https://phpstan.org/r/c7fa7bda-1024-45c3-957d-7b91de41b844)

herndlm:
[...]

PR #123, opened 2022-03-10
Add assertSame test case with static method

herndlm:
This is basically a regression test and closes https://github.com/phpstan/phpstan-phpunit/issues/120


herndlm:
Would be nice to have one for assertNotSame, which is actually used in that bug, too. mom

ondrejmirtes:
Thank you!

PR #130, opened 2022-05-04
Add rule that checks for invalid and unrecognized annotations

PrinsFrank:
When writing annotations, missing a space between the annotation and the value will result in the annotation not being recognized. This PR adds a rule that specific annotations that have a parameter should be followed by a space before that parameter.

### DIFF
--- a/src/Rules/PHPUnit/NoMissingSpaceInAnnotationRule.php
+++ b/src/Rules/PHPUnit/NoMissingSpaceInAnnotationRule.php
@@ -1,5 +1,4 @@
-<?php
-declare(strict_types=1);
+<?php declare(strict_types = 1);
 
 namespace PHPStan\Rules\PHPUnit;
 
@@ -7,12 +6,17 @@ use PhpParser\Node;
 use PHPStan\Analyser\Scope;
 use PHPStan\Rules\Rule;
 use PHPUnit\Framework\TestCase;
+use function in_array;
+use function is_subclass_of;
+use function preg_match;
+use function preg_split;
 
 /**
  * @implements Rule<Node>
  */
 class NoMissingSpaceInAnnotationRule implements Rule
 {
+
 	private const ANNOTATIONS_WITH_PARAMS = [
 		'backupGlobals',
 		'backupStaticAttributes',
@@ -26,7 +30,7 @@ class NoMissingSpaceInAnnotationRule implements Rule
 		'testDox',
 		'testWith',
 		'ticket',
-		'uses'
+		'uses',
 	];
 
 	public function getNodeType(): string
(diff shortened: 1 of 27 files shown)


PrinsFrank:
@ondrejmirtes any change this can get merged? Or is there anything I can improve?

ondrejmirtes:
Hi, I noticed your PR but I'm a bit torn on it because I've been a heavy PHPUnit user for the last 12 years at least, but I never used a single annotation like that (besides `@dataProvider` which I use a lot).

How severe are the bugs that are reported by this rule? Does it lead to something wrong being reported by PHPUnit, or is it that it plainly doesn't work and the user recognizes it easily?

PrinsFrank:
@ondrejmirtes for dataProviders the functionality breaks because a test with arguments doesn't work without a provider. That is not the case for all annotations. I've tried making this inspection generic, so I looked at the documentation for all annotations with arguments, but the one I actually created this rule for was the ```@covers``` annotation.

In open source packages, there is usually very tight control over the quality of code, but when you work in a bigger company on a monolith there is usually some variation of expertises between teams. To make sure unit tests are added for new features, at several of those projects we had a coverage that was not allowed to be decreased.

Sometimes though, a unit test was added that actually booted an entire framework or hit a lot of dependencies, causing the test coverage to be way higher than it should be for the unit tests because of the way coverage is measured. For this reason, we started to use ```@coversDefaultClass \F\Q\N``` above each test class and ```@covers ::methodName``` above each test method to make sure if extra code got executed it wouldn't impact the coverage. If you write it without the extra space, the annotation is not taken into consideration though: ```@covers::methodName```, which is really hard to spot in a code review.

ondrejmirtes:
This is really solid work, thanks! I'd like to merge it but for backward compatibility, the rule has to be only part of bleedingEdge: https://phpstan.org/blog/what-is-bleeding-edge

To do that, please follow this example: https://github.com/phpstan/phpstan-doctrine/blob/bc5036146b0badc1e07b450a6ae73086524cd953/rules.neon#L27-L29

PrinsFrank:
@ondrejmirtes Thanks for the feedback, I moved it to a bleedingEdge rule and made sure the branch was up to date again!

src/Rules/PHPUnit/NoMissingSpaceInAnnotationRule.php
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		if (!$node instanceof Node\Stmt\Class_ && !$node instanceof Node\Stmt\ClassMethod) {
+			return [];
+		}
+
+		if ($node instanceof Node\Stmt\Class_) {
+			if ($node->namespacedName !== null && $this->reflectionProvider->getClass($node->namespacedName->toString())->isSubclassOf(TestCase::class) === false) {
```
ondrejmirtes: I've looked at it again. These should be two separate rules. The best practice is to have a rule for `InClassNode` - you'll be able to get `$scope->getClassReflection()`. It's gonna work for anonymous classes, and also in some weird situations where the class might actually not exist.

And in case of methods you can use InClassMethodNode and ask for `$scope->getFunction()`. The nice thing about that is that you can then ask for `$scope->getFunction()->getDocComment()` too.

The common part between these two rules can be reused via some helper class that you'll inject into the constructor in both rules.

Thanks for understanding.

ondrejmirtes (changes requested):

src/Rules/PHPUnit/NoMissingSpaceInAnnotationRule.php
PrinsFrank: Thanks, I have updated the code to reflect these changes!

ondrejmirtes:
Thank you!

PR #137, opened 2022-07-24
Rules to check `@covers` and `@coversDefaultClass` for methods and classes

mad-briller:
continuation of https://github.com/phpstan/phpstan-phpunit/pull/119 but changed quite a bit

aswell as resolving the PR feedback in 119, i had to rework some of the code due to phpstan 1.6 deprecating the `parent` attribute, which meant major reworks on the rule itself to become 2 rules as per this discussion: https://github.com/phpstan/phpstan/discussions/7678

as a result this doesn't include any of the commits from the original author @neclimdul, as so much changed from the original PR it was easier to produce a new single commit with all the changes in.


neclimdul:
Ouch, those are clearly my tests. Also, looks like you lost one of the test cases.

```
/**
 * @covers \PHPUnit\Framework\TestCase
 */
class CoversShouldExistTestCase3 extends \PHPUnit\Framework\TestCase
{

	/**
	 * @covers \PHPUnit\Framework\TestCase
	 */
	public function testBadCoversDefault() {}

}
```

mad-briller:
@neclimdul yeah i started from your branch but then completely reworked it due to changes in phpstans core and the above discussion, it was not my intention to "take the credit" so to speak and that is why i mentioned you

this was my first time finishing somebody elses open source PR so i was not aware of the best etiquette, my apologies. i will rebase the work on top of yours now, and re-add the missing test case.

mad-briller:
i've rebased the work on your branch and also added a new error message when the `@covers` is redundant.

sorry about that, i hope you understand i had no ill intent :)

neclimdul:
Yeah, didn't expect you did. I don't have any intention of causing problems but seems like someone could be malicious with licensing if their code was committed as authored by someone else. Also just doesn't feel great to loose credit, but that's more a weakness in github only really capturing credit as the author.

src/Rules/PHPUnit/ClassCoversExistsRule.php
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	public function getNodeType(): string
+	{
+		return InClassNode::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		$classReflection = $scope->getClassReflection();
```
ondrejmirtes: You can get `$node->getClassReflection()` and save yourself a null check.

ondrejmirtes (changes requested):

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

ondrejmirtes:
Thank you.

PR #144, opened 2022-10-28
Fix covers rule for functions.

### LINKED ISSUE #143: @coverage annotation for function outside class
Hi!

With an update to `1.2.1` I get a new error reported regarding my `@coverage` annotation. The coverage is reported correctly for this function, however PHPStan tells me the annotation is incorrect.

image

Functions file:
```php
<?php

declare(strict_types=1);

namespace Domain\Shared\Translation;

/**
 * @param array<mixed> $parameters
 */
function t(string $message, array $parameters = [], string $domain = null): TranslatableMessage {}
```

Test class:

```php
<?php

declare(strict_types=1);

namespace Tests\Unit\Domain\Shared\Translation;

use Tests\Unit\UnitTestCase;

use function Domain\Shared\Translation\t;

/**
 * @covers \Domain\Shared\Translation\t
 */
class FunctionsTest extends UnitTestCase
{}
```

`composer.json`:
```
{
    "autoload": {
        "files": [
            "src/Domain/Shared/Translation/functions.php"
        ],
        "psr-4": {
            "Application\\": "src/Application",
            "Domain\\": "src/Domain",
            "Infrastructure\\": "src/Infrastructure",
            "UserInterface\\": "src/UserInterface"
        }
    },
}

### PR DESCRIPTION
**Pull request by mad-briller**
resolves: https://github.com/phpstan/phpstan-phpunit/issues/143

### DIFF
--- a/src/Rules/PHPUnit/CoversHelper.php
+++ b/src/Rules/PHPUnit/CoversHelper.php
@@ -72,6 +72,7 @@ class CoversHelper
 		$errors = [];
 		$covers = (string) $phpDocTag->value;
 		$isMethod = strpos($covers, '::') !== false;
+		$fullName = $covers;
 
 		if ($isMethod) {
 			[$className, $method] = explode('::', $covers);
@@ -81,6 +82,7 @@ class CoversHelper
 
 		if ($className === '' && $node instanceof Node\Stmt\ClassMethod && $coversDefaultClass !== null) {
 			$className = (string) $coversDefaultClass->value;
+			$fullName = $className . $covers;
 		}
 
 		if ($this->reflectionProvider->hasClass($className)) {
@@ -89,7 +91,7 @@ class CoversHelper
 			if (isset($method) && $method !== '' && !$class->hasMethod($method)) {
 				$errors[] = RuleErrorBuilder::message(sprintf(
 					'@covers value %s references an invalid method.',
-					$covers
+					$fullName
 				))->build();
 			}
 		} else {
@@ -99,7 +101,7 @@ class CoversHelper
 
 			$errors[] = RuleErrorBuilder::message(sprintf(
 				'@covers value %s references an invalid %s.',
-				$covers,
+				$fullName,
 				$isMethod ? 'method' : 'class or function'
 			))->build();
 		}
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: tests/Rules/PHPUnit/ClassMethodCoversExistsRuleTest.php [hunk 1]
```diff
@@ -26,15 +26,15 @@ public function testRule(): void
 	{
 		$this->analyse([__DIR__ . '/data/method-coverage.php'], [
 			[
-				'@covers value ::ignoreThis references an invalid class.',
+				'@covers value ::ignoreThis references an invalid method.',
```
mad-briller: these cases changed because the messaging was refined to indicate it's a method if `::` is found in the `@covers` annotation,

previously it would say `class` in all cases which was a little misleading

src/Rules/PHPUnit/CoversHelper.php
```diff
@@ -83,16 +85,22 @@ public function processCovers(
...
+			if (!isset($method) && $this->reflectionProvider->hasFunction(new Name($covers, []), null)) {
+				return $errors;
+			}
+
 			$errors[] = RuleErrorBuilder::message(sprintf(
-				'@covers value %s references an invalid class.',
-				$covers
+				'@covers value %s references an invalid %s.',
+				$covers,
+				$isMethod ? 'method' : 'class or function'
```
ondrejmirtes, line 105: I don't see any test for "class or function", can you find it or add it?

ondrejmirtes (changes requested):

src/Rules/PHPUnit/CoversHelper.php
mad-briller: good spot, i've added that now

tests/Rules/PHPUnit/ClassMethodCoversExistsRuleTest.php
ondrejmirtes: What would be great is to add the nonexistent names to all these messages. Be as explicit as possible 😊

mad-briller: as in the fully qualified name? i did consider adding that but didn't want to change too much at once!

ondrejmirtes: You can change it, it's in bleedingEdge 😊

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

darthf1:
Awesome, that's quick. Thank you @mad-briller and @ondrejmirtes

PR #150, opened 2022-12-07
Add rule to check @dataProvider

### LINKED ISSUE #148: Add a rule that checks for non-static data providers
Using a non-static method as a data provider will be deprecated from PHPUnit 10: https://github.com/sebastianbergmann/phpunit/commit/9caafe2d49b33a21f87db248a8ad6ca7c7bdac09

Adding a rule to detect them could be helpful.

villfa:
Here a new rule to check the `@dataProvider` annotation:
- does the method exist
- has the method the right case (depending on the `checkFunctionNameCase` parameter)
- is the method public
- ~is the method static~

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -55,6 +55,8 @@ services:
 		class: PHPStan\Rules\PHPUnit\CoversHelper
 	-
 		class: PHPStan\Rules\PHPUnit\AnnotationHelper
+	-
+		class: PHPStan\Rules\PHPUnit\DataProviderHelper
 
 conditionalTags:
 	PHPStan\PhpDoc\PHPUnit\MockObjectTypeNodeResolverExtension:
--- a/src/Rules/PHPUnit/DataProviderHelper.php
+++ b/src/Rules/PHPUnit/DataProviderHelper.php
@@ -11,7 +11,6 @@ use PHPStan\Rules\RuleErrorBuilder;
 use function array_merge;
 use function preg_match;
 use function sprintf;
-use function trim;
 
 class DataProviderHelper
 {
@@ -44,7 +43,8 @@ class DataProviderHelper
 	 */
 	public function processDataProvider(
 		Scope $scope,
-		PhpDocTagNode $phpDocTag
+		PhpDocTagNode $phpDocTag,
+		bool $checkFunctionNameCase
 	): array
 	{
 		$dataProviderName = $this->getDataProviderName($phpDocTag);
@@ -72,7 +72,7 @@ class DataProviderHelper
 
 		$errors = [];
 
-		if ($dataProviderName !== $dataProviderMethodReflection->getName()) {
+		if ($checkFunctionNameCase && $dataProviderName !== $dataProviderMethodReflection->getName()) {
 			$errors[] = RuleErrorBuilder::message(sprintf(
 				'@dataProvider %s related method is used with incorrect case: %s.',
 				$dataProviderName,
@@ -87,21 +87,12 @@ class DataProviderHelper
 			))->build();
 		}
 
-		if (!$dataProviderMethodReflection->isStatic()) {
-			$errors[] = RuleErrorBuilder::message(sprintf(
-				'@dataProvider %s related method must be static.',
-				$dataProviderName
-			))->build();
-		}
-
 		return $errors;
 	}
 
 	private function getDataProviderName(PhpDocTagNode $phpDocTag): ?string
 	{
-		$value = trim((string) $phpDocTag->value);
-
-		if (preg_match('/^[\S]+/', $value, $matches) !== 1) {
+		if (preg_match('/^[^ \t]+/', (string) $phpDocTag->value, $matches) !== 1) {
 			return null;
 		}
 
--- a/src/Rules/PHPUnit/DataProviderHelper.php
+++ b/src/Rules/PHPUnit/DataProviderHelper.php
@@ -1,0 +1,111 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\PHPUnit;
+
+use PHPStan\Analyser\Scope;
+use PHPStan\PhpDoc\ResolvedPhpDocBlock;
+use PHPStan\PhpDocParser\Ast\PhpDoc\PhpDocTagNode;
+use PHPStan\Reflection\MissingMethodFromReflectionException;
+use PHPStan\Rules\RuleError;
+use PHPStan\Rules\RuleErrorBuilder;
+use function array_merge;
+use function preg_match;
+use function sprintf;
+use function trim;
+
+class DataProviderHelper
+{
+
+	/**
+	 * @return array<PhpDocTagNode>
+	 */
+	public function getDataProviderAnnotations(?ResolvedPhpDocBlock $phpDoc): array
+	{
+		if ($phpDoc === null) {
+			return [];
+		}
+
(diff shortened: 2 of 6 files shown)


tests/Rules/PHPUnit/data/data-provider-declaration.php
```diff
@@ -0,0 +1,70 @@
...
+
+class FooTestCase extends \PHPUnit\Framework\TestCase
+{
+	/**
+	 * @dataProvider provideBar Comment.
+	 * @dataProvider providebaz
+	 * @dataProvider provideQux
+	 * @dataProvider provideQuux
+	 */
+	public function testIsNotFoo(string $subject): void
```
staabm, line 13: it would be great, if the rule could validate, whether the dataProviders return a correctly typed iterable in comparison to the expected parameter types of the test.

staabm (approved):
nice, thats something I had on my list for a few months :-).

the rule as is already provides a lot of value, so take my comment as an idea for further improvements - maybe in future PRs so we don't overcomplicate this very first iteration.

thanks!

extension.neon
```diff
@@ -55,6 +55,8 @@ services:
 		class: PHPStan\Rules\PHPUnit\CoversHelper
 	-
 		class: PHPStan\Rules\PHPUnit\AnnotationHelper
+	-
+		class: PHPStan\Rules\PHPUnit\DataProviderHelper
```
ondrejmirtes, line 59: The rule isn't registered in rules.neon. It needs to be enabled only for bleedingEdge, feel free to get inspired by other rules that are enabled like that.

src/Rules/PHPUnit/DataProviderHelper.php
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		$classReflection = $scope->getClassReflection();
+		if ($classReflection === null) {
+			// Should not happen
+			return [];
+		}
+
+		try {
+			$dataProviderMethodReflection = $classReflection->getMethod($dataProviderName, $scope);
```
ondrejmirtes: Needs to use `getNativeMethod`. I doubt that PHPUnit executes `__call`.

src/Rules/PHPUnit/DataProviderHelper.php
```diff
@@ -0,0 +1,111 @@
...
+			))->build();
+
+			return [$error];
+		}
+
+		$errors = [];
+
+		if ($dataProviderName !== $dataProviderMethodReflection->getName()) {
+			$errors[] = RuleErrorBuilder::message(sprintf(
+				'@dataProvider %s related method is used with incorrect case: %s.',
```
ondrejmirtes, line 77: This is too strict. I just checked and PHPUnit will also use dataProviders with wrong case. But you can still do it if you pass in the `checkFunctionNameCase` parameter.

src/Rules/PHPUnit/DataProviderHelper.php
```diff
@@ -0,0 +1,111 @@
...
+		if (!$dataProviderMethodReflection->isPublic()) {
+			$errors[] = RuleErrorBuilder::message(sprintf(
+				'@dataProvider %s related method must be public.',
+				$dataProviderName
+			))->build();
+		}
+
+		if (!$dataProviderMethodReflection->isStatic()) {
+			$errors[] = RuleErrorBuilder::message(sprintf(
+				'@dataProvider %s related method must be static.',
```
ondrejmirtes: This is too strict. Current version of PHPUnit doesn't require this, and for example all dataProviders in PHPStan are non-static.

src/Rules/PHPUnit/DataProviderHelper.php
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		return $errors;
+	}
+
+	private function getDataProviderName(PhpDocTagNode $phpDocTag): ?string
+	{
+		$value = trim((string) $phpDocTag->value);
+
+		if (preg_match('/^[\S]+/', $value, $matches) !== 1) {
```
ondrejmirtes: Does this match how the annotation is searched for in PHPUnit?

ondrejmirtes (changes requested):
Almost perfect! :) Just a few points to change.

src/Rules/PHPUnit/DataProviderHelper.php
villfa: I didn't think at comparing. Here what I've found:
- PHPUnit 9: https://github.com/sebastianbergmann/phpunit/blob/a4246d5/src/Util/Annotation/DocBlock.php#L68
  ```
  '/@dataProvider\s+([a-zA-Z0-9._:-\\\\x7f-\xff]+)/'
  ```
- PHPUnit 10: https://github.com/sebastianbergmann/phpunit/blob/9c0e694/src/Metadata/Parser/Annotation/DocBlock.php#L225
  ```
  '/@(?P<name>[A-Za-z_-]+)(?:[ \t]+(?P<value>.*?))?[ \t]*\r?$/m'
  ```

Which one do you suggest we need to match?

extension.neon
villfa: Should be ok now

src/Rules/PHPUnit/DataProviderHelper.php
villfa: changed

src/Rules/PHPUnit/DataProviderHelper.php
villfa: Well, that was the idea behind #148, to ease the migrations to the next version.

Would you be ok to make this configurable?

ondrejmirtes:
Please use the newer regex.

And about deprecations - let's revisit this after this PR is merged and released 👍

src/Rules/PHPUnit/DataProviderHelper.php
villfa: Done.
I hope I have the configuration part right.

src/Rules/PHPUnit/DataProviderHelper.php
villfa: Removed for now

src/Rules/PHPUnit/DataProviderHelper.php
villfa: Changed to match with the newer regex

ondrejmirtes:
Thank you, it's perfect!

About the next steps: I'm thinking the deprecation should be reported by phpstan-phpunit when also phpstan-deprecation-rules are installed.

These steps need to be taken:

1) Introduce new `deprecationRulesInstalled` parameter in PHPStan core and set it to `false`.
2) Require `phpstan/phpstan: ^1.9.3` in phpstan-deprecation-rules and set this parameter to `true`.
3) Require `phpstan/phpstan: ^1.9.3` in phpstan-phpunit and use the `deprecationRulesInstalled` parameter to trigger the wanted behaviour.

Can you execute this plan? :) Thank you.

villfa:
> Can you execute this plan?

Sure :)

tests/Rules/PHPUnit/data/data-provider-declaration.php
staabm: implemented it in https://github.com/phpstan/phpstan-phpunit/pull/152

ondrejmirtes:
Looks like PHPUnit 10 is going to also support attributes for dataProviders, can you please updste the rule so that it reads them too? Thanks!

PR #151, opened 2022-12-07
Report data providers deprecated usage

### LINKED ISSUE #148: Add a rule that checks for non-static data providers

villfa:
Last part of the plan described here https://github.com/phpstan/phpstan-phpunit/pull/150#issuecomment-1341164151

Closes #148

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
 	],
 	"require": {
 		"php": "^7.2 || ^8.0",
-		"phpstan/phpstan": "^1.9.0"
+		"phpstan/phpstan": "^1.9.3"
 	},
 	"conflict": {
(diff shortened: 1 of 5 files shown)


ondrejmirtes:
Perfect, thank you 😊 We have to wait until 1.9.3 is released and then I'll release these packages too.

1.9.x isn't currently in a releasable state because of some small logic and performance regressions, but the release shouldn't be too far off.

PR #152, opened 2022-12-07
validate dataProvider return/parameter types

### LINKED ISSUE #70: typecheck dataproviders
it would be great when phpstan could typecheck a dataprovider function against the type-signature of the methods which is using it.

see this code taken from composer/semver

the dataprovider will never provide a `null` value for `$min`.

```
    /**
     * @dataProvider hyphenConstraints
     *
     * @param string     $input
     * @param Constraint|null $min
     * @param Constraint $max
     */
    public function testParseHyphen($input, $min, $max)
    {
        $parser = new VersionParser();
        if ($min) {
            $expected = new MultiConstraint(array($min, $max));
        } else {
            $expected = $max;
        }

        $this->assertSame((string) $expected, (string) $parser->parseConstraints($input));
    }

    /**
     * @return array
     */
    public function hyphenConstraints()
    {
        return array(
            array('v1 - v2', new Constraint('>=', '1.0.0.0-dev'), new Constraint('<', '3.0.0.0-dev')),
            array('1.2.3 - 2.3.4.5', new Constraint('>=', '1.2.3.0-dev'), new Constraint('<=', '2.3.4.5')),
            array('1.2-beta - 2.3', new Constraint('>=', '1.2.0.0-beta'), new Constraint('<', '2.4.0.0-dev')),
            array('1.2-beta - 2.3-dev', new Constraint('>=', '1.2.0.0-beta'), new Constraint('<=', '2.3.0.0-dev')),
            array('1.2-RC - 2.3.1', new Constraint('>=', '1.2.0.0-RC'), new Constraint('<=', '2.3.1.0')),
            array('1.2.3-alpha - 2.3-RC', new Constraint('>=', '1.2.3.0-alpha'), new Constraint('<=', '2.3.0.0-RC')),
            array('1 - 2.0', new Constraint('>=', '1.0.0.0-dev'), new Constraint('<', '2.1.0.0-dev')),
            array('1 - 2.1', new Constraint('>=', '1.0.0.0-dev'), new Constraint('<', '2.2.0.0-dev')),
            array('1.2 - 2.1.0', new Constraint('>=', '1.2.0.0-dev'), new Constraint('<=', '2.1.0.0')),
            array('1.3 - 2.1.3', new Constraint('>=', '1.3.0.0-dev'), new Constraint('<=', '2.1.3.0')),
        );
    }
```

staabm:
Closes https://github.com/phpstan/phpstan-phpunit/issues/70

### DIFF
--- a/src/Rules/PHPUnit/DataProviderDeclarationRule.php
+++ b/src/Rules/PHPUnit/DataProviderDeclarationRule.php
@@ -84,6 +84,11 @@ class DataProviderDeclarationRule implements Rule
 
 		$annotations = $this->dataProviderHelper->getDataProviderAnnotations($methodPhpDoc);
 
+		$testMethodReflection = null;
+		if ($classReflection->hasMethod($node->name->toString())) {
+			$testMethodReflection = $classReflection->getMethod($node->name->toString(), $scope);
+		}
+
 		$errors = [];
 
 		foreach ($annotations as $annotation) {
@@ -93,7 +98,8 @@ class DataProviderDeclarationRule implements Rule
 					$scope,
 					$annotation,
 					$this->checkFunctionNameCase,
-					$this->deprecationRulesInstalled
+					$this->deprecationRulesInstalled,
+					$testMethodReflection
 				)
 			);
 		}
--- a/src/Rules/PHPUnit/DataProviderHelper.php
+++ b/src/Rules/PHPUnit/DataProviderHelper.php
@@ -5,10 +5,15 @@ namespace PHPStan\Rules\PHPUnit;
 use PHPStan\Analyser\Scope;
 use PHPStan\PhpDoc\ResolvedPhpDocBlock;
 use PHPStan\PhpDocParser\Ast\PhpDoc\PhpDocTagNode;
+use PHPStan\Reflection\ExtendedMethodReflection;
 use PHPStan\Reflection\MissingMethodFromReflectionException;
+use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Rules\RuleError;
 use PHPStan\Rules\RuleErrorBuilder;
+use PHPStan\Type\UnionType;
+use PHPStan\Type\VerbosityLevel;
 use function array_merge;
+use function count;
 use function preg_match;
 use function sprintf;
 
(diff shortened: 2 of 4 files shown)


ondrejmirtes:
I'd like a completely different approach than this.

1) Data providers usually don't come with precise typehints. So instead of comparing the return type of a data provider against the test case method signature, I'd like to inspect `return`, `yield`, and `yield from` in data providers against the methods that expects them.
2) I don't want to reimplement checking types passed into methods. PHPStan itself already does this, and contains years of edge case fixes. But it's possible that the logic (https://github.com/phpstan/phpstan-src/blob/1.9.x/src/Rules/FunctionCallParametersCheck.php) isn't reusable the best way so maybe first we need to refactor it.

villfa:
It may be important to note that `@dataProvider` can be used in combination with `@depends`, and it affects the tests signature, see [documentation](https://phpunit.readthedocs.io/en/9.5/writing-tests-for-phpunit.html#writing-tests-for-phpunit-data-providers-examples-dependencyanddataprovidercombo-php).

mvorisek:
> 1. Data providers usually don't come with precise typehints. So instead of comparing the return type of a data provider against the test case method signature, I'd like to inspect `return`, `yield`, and `yield from` in data providers against the methods that expects them.

I agree. With more and more complex method types, sometimes even conditional, it makes sense to actually not have any return type above data provider methods and infer the type from what the tests that require them expect.

stof:
Well, I would say that if the data provider method defines its return type, it should be used by the check instead of inferring it (as other rules will validate that the returned values match this type). But inferring it automatically when the type is not defined (or is just `iterable` or `array` without specifying the iterable value type), inferring is a good idea.

mvorisek:
> Well, I would say that if the data provider method defines its return type, it should be used by the check instead of inferring it (as other rules will validate that the returned values match this type). But inferring it automatically when the type is not defined (or is just `iterable` or `array` without specifying the iterable value type), inferring is a good idea.

Infered/accepting type (defined by accepting test method) should be always used.

However, data provider type might be narrower, the infered/accepting type should be intersected with this type.

staabm:
I will re-evaluate based on the given feedback and start a new PR if I have something to review

PR #155, opened 2022-12-11
Update .gitattributes

kubawerlos:
This PR:

 - adds a leading slash to all `export-ignore` entries
 - adds `/.*` which will match `.editorconfig` that previously was not excluded
 - excludes `build-cs` directory

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,13 +2,12 @@
 *.stub linguist-language=PHP
 *.neon linguist-language=YAML
 
-.github export-ignore
-tests export-ignore
-tmp export-ignore
-.gitattributes export-ignore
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thank you!

BTW If you could find a way to fix .gitattributes in https://github.com/phpstan/phpstan so that the major language isn't detected as Nunjucks (from `/website`), that'd be great!

I tried to use `linguist-vendored` based on https://github.com/github/linguist/blob/master/docs/overrides.md but without luck.

kubawerlos:
@ondrejmirtes [this](https://github.com/phpstan/phpstan/pull/8507) should work

PR #153, opened 2022-12-11
Discover data providers from other classes

kubawerlos:
Until 1.2.2 such code (working well) was correctly not reported:
```php
<?php
namespace Tests;
class FooTest extends TestCase
{
    public static function fooProvider(): iterable
    {
        /* ... */
    }
}
```
```php
<?php
namespace Tests;
class BarTest extends TestCase
{
    /**
     * @dataProvider \Tests\FooTest::fooProvider
     */
    public function testStuff(... $arguments): void
    {
        /* ... */
    }
}
````

(see failing job at https://github.com/kubawerlos/php-cs-fixer-custom-fixers/pull/849)

PR #156, opened 2022-12-12
Discover data providers from other classes

villfa:
Fix for the issue highlighted by #153.

### DIFF
--- a/src/Rules/PHPUnit/DataProviderHelper.php
+++ b/src/Rules/PHPUnit/DataProviderHelper.php
@@ -5,15 +5,31 @@ namespace PHPStan\Rules\PHPUnit;
 use PHPStan\Analyser\Scope;
 use PHPStan\PhpDoc\ResolvedPhpDocBlock;
 use PHPStan\PhpDocParser\Ast\PhpDoc\PhpDocTagNode;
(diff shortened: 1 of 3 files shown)


src/Rules/PHPUnit/DataProviderHelper.php
```diff
@@ -5,16 +5,32 @@
...
+	/**
+	 * Reflection provider.
+	 *
+	 * @var ReflectionProvider
+	 */
+	private $reflectionProvider;
+
+	public function __construct(ReflectionProvider $reflectionProvider)
+	{
+		$this->reflectionProvider = $reflectionProvider;
```
kubawerlos, line 31: So there is auto wiring it the project? I wanted to fix it by myself, but failed on getting reflection of other class and it is that simple :)

kubawerlos (approved):
👍🏼

src/Rules/PHPUnit/DataProviderHelper.php
ondrejmirtes: https://phpstan.org/developing-extensions/dependency-injection-configuration#services

ondrejmirtes:
Thank you. I can release this after PHPStan 1.9.3 is released.

PR #159, opened 2022-12-21
Add support for data provider attributes

villfa:
Add support for data provider attributes as suggested there: https://github.com/phpstan/phpstan-phpunit/pull/150#issuecomment-1344025037

### DIFF
--- a/src/Rules/PHPUnit/DataProviderDeclarationRule.php
+++ b/src/Rules/PHPUnit/DataProviderDeclarationRule.php
@@ -5,7 +5,6 @@ namespace PHPStan\Rules\PHPUnit;
 use PhpParser\Node;
 use PHPStan\Analyser\Scope;
 use PHPStan\Rules\Rule;
-use PHPStan\Type\FileTypeMapper;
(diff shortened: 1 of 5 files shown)


src/Rules/PHPUnit/DataProviderDeclarationRule.php
```diff
@@ -61,6 +51,9 @@ public function getNodeType(): string
 		return Node\Stmt\ClassMethod::class;
 	}
 
+	/**
+	 * @param Node\Stmt\ClassMethod $node
```
ondrejmirtes: This isn't needed because there's already `@implements Rule<Node\Stmt\ClassMethod>` above the class. Latest PhpStorm already understands this.

ondrejmirtes (changes requested):

src/Rules/PHPUnit/DataProviderDeclarationRule.php
villfa: Indeed. Removed.
Thanks

ondrejmirtes:
Thank you!

PR #166, opened 2023-02-08
Fixed deprecations in phpstan.

mad-briller:


phpstan-baseline.neon
```diff
@@ -1,5 +1,10 @@
 parameters:
 	ignoreErrors:
+		-
+			message: "#^Call to an undefined method PHPStan\\\\Type\\\\Type\\:\\:getClassName\\(\\)\\.$#"
```
ondrejmirtes, line 4: You can't do this, it's not safe obviously.

src/Rules/PHPUnit/MockMethodCallRule.php
```diff
@@ -81,7 +80,7 @@ public function processNode(Node $node, Scope $scope): array
 
 			$mockClass = $type->getTypes()[0];
 
-			if (!($mockClass instanceof ObjectType) || $mockClass->hasMethod($method)->yes()) {
+			if (!$mockClass->isObject()->yes() || $mockClass->hasMethod($method)->yes()) {
```
ondrejmirtes, line 83: The thing with `GenericObjectType` above can be rewritten using getTemplateType (https://apiref.phpstan.org/1.10.x/source-src.Type.Type.html#209-225): `$type->getTemplateType(InvocationMocker::class, 'TMockedClass')`

The part with `$mockClass->getClassName()` should be replaced with `$mockClass->getObjectClassNames()`.

ondrejmirtes (changes requested):

phpstan-baseline.neon
mad-briller: ah ok, i thought it was caused by removing the `instanceof` meaning that phpstan wasn't aware that the `isObject()->yes()` meant that function was available

ondrejmirtes: But it doesn't and shouldn't make that method available. Type is still Type.

mad-briller:
resolved in https://github.com/phpstan/phpstan-phpunit/commit/db436df51b0de4301fd1b522f9c7aa455d8427b8
thanks ondrej

PR #168, opened 2023-02-23
Add tip to error when a not fully qualified name is seen in @covers annotation.

mad-briller:
This is a mistake i see made fairly often and is only raised as an issue by phpunit when actually generating coverage reports, which is often only done in ci and not when running locally, so would be great to give the feedback sooner.


ondrejmirtes:
Thank you!

PR #169, opened 2023-02-28
Fixed false positive when covering a global function.

mad-briller:
Previously this case wasn't tested and was overlooked as phpstan test data files are namespaced, but this cropped up today during rollout of the rules and so i've had to `@covers` a php built-in to cover the case in a test

also moved the "doesn't specify anything" test up in the code as it's a fail-fast path, dunno why i put it where i did in the first place, time pressure probably

thanks for your time


ondrejmirtes:
Thank you!

PR #174, opened 2023-03-24
Warn when trying to cover an interface.

mad-briller:
phpunit 10.x errors when attempting to `@covers` an interface, so phpstan can warn about that case early


ondrejmirtes:
Thank you!

PR #180, opened 2023-05-02
Add partial support for assertContainsOnlyInstancesOf

### LINKED ISSUE #158: PhpStan does not consider usage of Phpunit Assert::assertNotEmpty and Assert::assertContainsOnlyInstancesOf when returning an Array
Hello,

I am building a test returning an array of entities to be used in some others tests marked with `@depends`.
Despite I check that my returned array is not empty an contains only instances of a specific class, phpstan keeps alerting
that my returned value does not match. Here is my code: 

```php
     /**
     * @return Post[]
     *
     * @throws \Exception
     */
    public function testFindForHomePageReturnsCollectionOfPosts(): array
    {
        $testSubject = self::getContainer()->get(PostRepository::class);
        if (!$testSubject instanceof PostRepository) {
            throw new ServiceNotFoundException(PostRepository::class);
        }

        $result = $testSubject->findForHomePage()->toArray();
        self::assertNotEmpty($result);
        self::assertContainsOnlyInstancesOf(Post::class, $result);

        return $result;
    }
```

And the alert thrown by phpstan (line 34 is the `return $result` statement): 
image

Context: 
---
Symfony 6.2 KernelTestCase
phpunit/phpunit 9.5.25
symfony/phpunit-bridge 6.2.0

PhpStan 1.9.3
extensions:
 - phpstan-phpunit 1.3.2
 - phpstan-doctrine 1.3.18
 - phpstan-symfony 1.2.15
 - phpstan/phpstan-webmozart-assert 1.2.1

Thanks a lot for your help :)

villfa:
Add support for [assertContainsOnlyInstancesOf](https://docs.phpunit.de/en/9.5/assertions.html#assertcontainsonlyinstancesof) when the 2nd argument is an array.
The behaviour is not changed when it is an instance of *Traversable*.

See #158

### DIFF
--- a/src/Type/PHPUnit/Assert/AssertTypeSpecifyingExtensionHelper.php
+++ b/src/Type/PHPUnit/Assert/AssertTypeSpecifyingExtensionHelper.php
@@ -13,7 +13,9 @@ use PhpParser\Node\Expr\ConstFetch;
 use PhpParser\Node\Expr\FuncCall;
 use PhpParser\Node\Expr\Instanceof_;
 use PhpParser\Node\Name;
+use PhpParser\Node\Param;
 use PhpParser\Node\Scalar\LNumber;
+use PhpParser\Node\Stmt;
 use PHPStan\Analyser\Scope;
 use PHPStan\Analyser\SpecifiedTypes;
 use PHPStan\Analyser\TypeSpecifier;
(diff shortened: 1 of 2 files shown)


villfa:
~I move this PR back to draft since I forgot to consider the argument could be an instance of *Traversable*...~
**EDIT**: I've added a test to be sure there is no regression when the 2nd argument is an instance of *Traversable*, in order to make this PR ready to be reviewed.

ondrejmirtes:
Thank you.

Issue #167, opened 2023-02-10
ArrayObject will always evaluate to false


altdovydas:
# Bug report

Tested with  1.9.11 - 1.9.17
PHP 8.1.10
PHPUnit 9.5.27

### Code snippet that reproduces the problem

```
$arrayObject = new ArrayObject(['test' => true]);
$this->assertArrayHasKey('test', $arrayObject);
```
### Result
`
Call to method PHPUnit\Framework\Assert::assertArrayHasKey() with 'test' and ArrayObject<string, bool> will always evaluate to false.
`

### Expected output

OK

### Code snippet with god results
```
$arrayObject = new ArrayObject(['test']);
$this->assertArrayHasKey('test', $arrayObject);
```

#### PHPUnit output:
`
Failed asserting that an array has the key 'test'.
`
#### PHPStan output:
`
Call to method PHPUnit\Framework\Assert::assertArrayHasKey() with 'test' and ArrayObject<int, string> will always evaluate to false.
`

derrabus:
This issue seems to be fixed in 1.3.14 while at the same time false positives are now reported for `assertNotContains()`.

derrabus:
With 1.3.15, I cannot reproduce either of the issues. `assertArrayHasKey()` and `assertNotContains()` both seem to be understood correctly now.

ondrejmirtes:
Great, thank you :)

PR #182, opened 2023-05-12
Improve support of assertArrayHasKey()

villfa:
Related to #167

### DIFF
--- a/tests/Type/PHPUnit/data/assert-function.php
+++ b/tests/Type/PHPUnit/data/assert-function.php
@@ -31,15 +31,21 @@ class Foo
 		assertType(\DateTimeInterface::class, $o);
 	}
 
-	public function arrayHasNumericKey(array $a): void {
+	public function arrayHasNumericKey(array $a, \ArrayAccess $b): void {
 		assertArrayHasKey(0, $a);
 		assertType('array&hasOffset(0)', $a);
+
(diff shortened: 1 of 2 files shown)


tests/Type/PHPUnit/data/assert-function.php
```diff
@@ -20,15 +20,21 @@ public function doFoo($o): void
 		assertType(self::class, $o);
 	}
 
-	public function arrayHasNumericKey(array $a): void {
+	public function arrayHasNumericKey(array $a, \ArrayAccess $b): void {
 		assertArrayHasKey(0, $a);
 		assertType('array&hasOffset(0)', $a);
+
+		assertArrayHasKey(0, $b);
+		assertType('ArrayAccess', $b);
```
villfa, line 39: I wanted to obtain this result but for some reasons it didn't work. However before my changes the result was `*NEVER*` so it's already an improvement.
```suggestion
		assertType('ArrayAccess&hasOffset(0)', $b);
```

ondrejmirtes:
Sorry for the late review - please rebase because of a conflict. Thank you.

As for your suggestion - HasOffsetType and HasOffsetValueType support only intersection with arrays.

villfa:
The branch is now up to date.

ondrejmirtes:
Thank you!

Issue #185, opened 2023-05-22
assertInstanceOf doesn't refine type when used with generic class-string


Jean85:
I want to implement this method in my testcase:

```php
class BaseTestCase extends \PHPUnit\Framework\TestCase
{
    /**
     * @template T of object
     *
     * @param class-string<T> $class
     *
     * @return T
     */
    protected function fetch(string $class): object
    {
        $message = $this->fetch(); // returns object of unknown class

        $this->assertInstanceOf($class, $message, 'Message is not of the expected type');

        return $firstMessage;
    }
}
```

but this doesn't seem to work.

Jean85:
I hope to have recreated a correct regression test for this in #186

ondrejmirtes:
Fixed https://github.com/phpstan/phpstan-phpunit/pull/186

PR #186, opened 2023-05-22
Add regression test for #185

Jean85:


tests/Type/PHPUnit/data/assert-function.php
```diff
@@ -20,6 +20,17 @@ public function doFoo($o): void
...
 
+	/**
+	 * @template T of object
+	 * @param object $o
+	 * @param class-string<T> $class
+	 */
+	public function assertInstanceOfWorksWithTemplate($o, $class): void
+	{
+		assertInstanceOf($class, $o);
+		assertType($class, $o);
```
ondrejmirtes: This isn't a valid test case. The first parameter of `assertType` needs to be a literal string like `'Foo'`.

ondrejmirtes (changes requested):

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

PR #189, opened 2023-05-26
Microoptimize AssertRuleHelper.

mad-briller:
minor micro-opts to AssertRuleHelper

this static method is called multiple times for every ast node analysed currently due to the rules that use it specifying `NodeAbstract` as their node type


mad-briller:
not sure why the actions are failing, just says "failed to setup phpX.X"

ondrejmirtes:
Yeah, that's not your fault. Thank you!

PR #191, opened 2023-07-21
Add support for assertContains and assertContainsEquals

villfa:
The implementation for `assertContainsEquals()` is slightly more complicated because of [a problem with PHPStan](https://github.com/phpstan/phpstan/issues/9662).

See also [PHPUnit documentation](https://docs.phpunit.de/en/10.0/assertions.html#assertcontains)


ondrejmirtes:
Thank you!

PR #192, opened 2023-08-23
Add support for `assertObjectHasProperty`

Wirone:
[Added in PHPUnit 9.6.11](https://github.com/sebastianbergmann/phpunit/blob/9.6/ChangeLog-9.6.md#9611---2023-08-19) as a forward-compatibility.

### DIFF
--- a/tests/Type/PHPUnit/AssertFunctionTypeSpecifyingExtensionTest.php
+++ b/tests/Type/PHPUnit/AssertFunctionTypeSpecifyingExtensionTest.php
@@ -11,17 +11,15 @@ class AssertFunctionTypeSpecifyingExtensionTest extends TypeInferenceTestCase
 	/** @return mixed[] */
 	public function dataFileAsserts(): iterable
 	{
-		if (!function_exists('PHPUnit\\Framework\\assertInstanceOf')) {
-			return [];
+		if (function_exists('PHPUnit\\Framework\\assertInstanceOf')) {
+			yield from $this->gatherAssertTypes(__DIR__ . '/data/assert-function.php');
 		}
 
-		yield from $this->gatherAssertTypes(__DIR__ . '/data/assert-function.php');
-
-		if (!function_exists('PHPUnit\\Framework\\assertObjectHasProperty')) {
-			return [];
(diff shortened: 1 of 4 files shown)


Wirone:
@ondrejmirtes should I bump minimum PHPUnit version to 9.6.11 in order to satisfy CI? Because these functions are not available in 9.5 which is currently required.

Wirone:
For reference: it should be handle like [this](https://github.com/phpstan/phpstan-doctrine/blob/bf9c2ea1e1f82e9671754c6df6692a72111e0427/tests/DoctrineIntegration/ORM/EntityManagerTypeInferenceTest.php#L22-L23), with conditional assert based on PHPUnit version installed.

tests/Type/PHPUnit/AssertFunctionTypeSpecifyingExtensionTest.php
```diff
@@ -16,6 +16,12 @@ public function dataFileAsserts(): iterable
 		}
 
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/assert-function.php');
+
+		if (!function_exists('PHPUnit\\Framework\\assertObjectHasProperty')) {
+			return [];
+		}
+
+		yield from $this->gatherAssertTypes(__DIR__ . '/data/assert-function-9.6.11.php');
```
ondrejmirtes: I don't like this structure. I know PHPCS requires it, but I like this more:

```php
		something();
		if (function_exists('PHPUnit\\Framework\\assertObjectHasProperty')) {
			yield from $this->gatherAssertTypes(__DIR__ . '/data/assert-function-9.6.11.php');
		}
		something();
```

You can change the order of other `yield from ...` calls in this method.

ondrejmirtes (changes requested):

tests/Type/PHPUnit/AssertFunctionTypeSpecifyingExtensionTest.php
Wirone: Yeah, I initially did it differently, but changed to satisfy PHPCS. I'll look at it later, AFK at this moment.

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

PR #197, opened 2023-10-09
AssertTypeSpecifyingExtensionHelper: rootExpr with unknown variable to avoid always-true false positives

### LINKED ISSUE #193: `assertContains` always claiming it'll return true
Reproduction case:

```php
        $this->assertContains('not in the list', new \ArrayObject([1]));
```

This generates the error:
> Call to method PHPUnit\Framework\Assert::assertContains() with 'not in the list' and ArrayObject<int, int> will always evaluate to true.

This is obviously incorrect and appears to happen whenever `$haystack` is a `\Traversable` rather than an array.

janedbal:
This should fix https://github.com/phpstan/phpstan-phpunit/issues/193.


janedbal:
Optionally, we might do this hack only for those resolvers where false-positive is at risk (like the assertContains with Traversable).

Edit: done, it breaks stuff without that.

ondrejmirtes:
Thank you!

PR #205, opened 2024-02-23
Narrow type on assertCount and assertNotCount

### LINKED ISSUE #204: Narrow type on assertCount to non-empty-*
As noted here https://github.com/phpstan/phpstan-phpunit/issues/54#issuecomment-946069516

```php
Assert::assertCount(1, $array);
```

Not sure if it is possible to implement narrowing type to non-empty-* here.
This would be very helpful, so we do not need to add assertNotEmpty assertion.

Thanks

michalbundyra:
Fixes #204

### DIFF
--- a/tests/Type/PHPUnit/data/assert-function.php
+++ b/tests/Type/PHPUnit/data/assert-function.php
@@ -7,6 +7,8 @@ use function PHPUnit\Framework\assertArrayHasKey;
 use function PHPUnit\Framework\assertContains;
 use function PHPUnit\Framework\assertContainsEquals;
 use function PHPUnit\Framework\assertContainsOnlyInstancesOf;
+use function PHPUnit\Framework\assertCount;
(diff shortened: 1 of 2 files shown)


tests/Type/PHPUnit/data/assert-function.php
```diff
@@ -90,4 +92,23 @@ public function containsOnlyInstancesOf(array $a, \Traversable $b): void
 		assertType('Traversable', $b);
 	}
 
+	public function count(array $a, \Countable $b): void
+	{
+		assertCount(3, $a);
+		assertType('non-empty-array', $a);
+
+		assertCount(7, $b);
+		assertType('Countable', $b);
```
michalbundyra, line 101: not sure if we have any better type for non-empty Countable ?

ondrejmirtes:
Thank you.

PR #207, opened 2024-04-04
Addition of rule for cases better represented with assertSameSize() method

o0h:
This Pull Request enhances the AssertSameWithCountRule in PHPUnit to handle additional use cases. The new rule now checks when both sides of the assertSame are counts, and suggests assertSameSize instead when necessary.

Here is some detail on what has changed:
* AssertSameWithCountRule now includes a check for cases where expected value of assertCount is the direct result of count function call.
* In these cases, a special assertion function assertSameSize is suggested to compare the sizes of collections rather than using the original assertSame function.

The code changes involve adjustments in method call structures, evaluation of counts and introduction of cases for assertSameSize.

### DIFF
--- a/src/Rules/PHPUnit/AssertSameWithCountRule.php
+++ b/src/Rules/PHPUnit/AssertSameWithCountRule.php
@@ -4,6 +4,7 @@ namespace PHPStan\Rules\PHPUnit;
 
 use Countable;
 use PhpParser\Node;
+use PhpParser\Node\Expr\MethodCall;
 use PhpParser\NodeAbstract;
 use PHPStan\Analyser\Scope;
 use PHPStan\Rules\Rule;
@@ -16,6 +17,14 @@ use function count;
  */
 class AssertSameWithCountRule implements Rule
 {
+
+	/** @var bool */
+	private $bleedingEdge;
+
+	public function __construct(bool $bleedingEdge)
+	{
+		$this->bleedingEdge = $bleedingEdge;
+	}
 
 	public function getNodeType(): string
 	{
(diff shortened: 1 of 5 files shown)


ondrejmirtes (changes requested):
Hi,
this should only be reported with bleeding edge. Use `private bool $bleedingEdge` in the rule constructor and add the new behaviour only when this is true. Here's an example how it should look in config: https://github.com/phpstan/phpstan-doctrine/blob/0d967379b8ea58c4f4499b17eeb9f29718350497/rules.neon#L57

o0h:
Hi, @ondrejmirtes!

Thanks for the quick feedback! 
Sorry that it's not passing the CI checks just yet. I'll sort out the issue and let you know once it's ready for review.

I'll be sure to use the bleeding edge flag as you suggested. Just hang tight a little longer!

Best regards,

o0h:
I have completed the necessary corrections that were of concern.
I would appreciate it if you could take a look and check it.
Thank you 😃

src/Rules/PHPUnit/AssertSameWithCountRule.php
```diff
@@ -37,36 +46,89 @@ public function processNode(Node $node, Scope $scope): array
...
-					->build(),
-			];
+		$rightIsCountFuncCall = $this->isCountFuncCall($right);
+		$rightIsCountMethodCall = $this->isCountMethodCall($right) && $this->argIsCountable($right, $scope);
+		if (!($rightIsCountFuncCall || $rightIsCountMethodCall)) {
+			return [];
 		}
 
-		if (
-			$right instanceof Node\Expr\MethodCall
```
ondrejmirtes, line 53: TBH this diff is very messy. I can't tell if the no-bleedingEdge behaviour is preserved or not. Can you split the changes in multiple commits? First extract logic into `isCount...` methods without changing the behaviour, and then introduce `bleedingEdge-on` behaviour that will be a new separate `if` statement. Can you please do that? Thank you.

ondrejmirtes (changes requested):

Issue #212, opened 2024-11-12
After `Assert::assertIsList` variable is not a list


mitelg:
Hey there :wave: 

I recently found that, the assertion `\PHPUnit\Framework\Assert::assertIsList` is not handled correctly. The variable has the following type after this check `Dumped type: array<int, string>`. But it should be `Dumped type: list<string>` instead. 

If you would point me to the correct please, I could try to provide a fix. 

Best regards

ondrejmirtes:
There's no code around assertIsList in this extension so PHPStan solely relies on the PHPDoc above the function which looks like this: https://github.com/sebastianbergmann/phpunit/blob/d81cfde3082807ab0db366c8424cb5bbbfcdc03b/src/Framework/Assert.php#L196-L206

So it does not provide any assert.

You could send a PR to PHPUnit so that this PHPDoc is added. Meanwhile it could be fixed by adding a stub in phpstan-phpunit.

mitelg:
I am not sure, if I understand correctly, what needs to be done :see_no_evil: 

You mean only by adding a comment like `Asserts that an array is a list.` above the method is enough? :thinking: Other than that, I do not really see a difference to other methods

ondrejmirtes:
You need to add `@phpstan-assert` PHPDoc tag above the method. PHPUnit has that already in many places.

mitelg:
ah! :bulb: I was only looking at the methods around `assertIsList` and was wondering how this should work :sweat_smile: 

thanks, I will provide a fix for PHPUnit then :+1:

PR #213, opened 2024-11-12
fix: Make sure Assert::assertIsList correctly asserts an array as list

mitelg:
Follow up of this issue: https://github.com/phpstan/phpstan-phpunit/issues/212

I decided to also bring this up into this repo, to make sure it also works for older PHPUnit versions. 

for PHPUnit 11 it is fixed here: https://github.com/sebastianbergmann/phpunit/pull/6027

but the `phpstan-assert` annotations are not used in older versions. I hope this is okay.

I also add it to the 1.4.x branch, is we could not update to 2.x yet, because of missing plugin compatibility.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -9,6 +9,9 @@ parameters:
 			- markTestIncomplete
 			- markTestSkipped
 	stubFiles:
+		- stubs/Assert.stub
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
Thank you.

Issue #214, opened 2024-11-26
Rule to enforce class naming


cs278:
Generally speaking PHPUnit looks for tests in files named `*Test.php`, which conventionally means the classes should be named `*Test`. Additionally I believe it is recommended/required (I cannot find docs for this) that "tests" designed to be extended should be named `*TestCase`.

I propose adding rule to this extension which enforces that classes are named appropriately:

* If a concrete class extends `PHPUnit\Framework\TestCase` and it should be named `*Test` - could also enforce `final`?
* If an abstract class extends `PHPUnit\Framework\TestCase` it should be named `*TestCase`.

Ideally the rule would also enforce the file is named appropriately as well but that might be outside scope.

If this convention isn't followed and you don't explicitly tell PHPUnit your test files then tests are not run, potentially keeping the build passing because a number of tests are being silently ignored.

ruudk:
I like this idea.

ruudk:
Another one: all public non-static methods should have a `#[Test]` attribute or named like `test*`. It happened a few times we thought we had tests but they were not picked up.

ruudk:
This could be a good starting point for the final / abstract part, or as inspiration: https://github.com/ergebnis/phpstan-rules/blob/main/src/Classes/FinalRule.php

cs278:
@ruudk I've thrown a PR together implementing my proposal, your proposal for checking methods as well could be added as an additional rule.

staabm:
since I also stumbled recently over the "name your abstract test class" with a "TestCase" suffix, I really like this proposal.

ondrejmirtes:
Hey, this isn't generally applicable. The way PHPUnit looks for test files is configurable. It's just that it usually looks like this:

```
<directory suffix="Test.php">
```

I feel like these suggested rules are too opinionated for the general audience.

cs278:
Damn, I'd completely forgotten this was even configurable.

PR #215, opened 2024-11-27
Add rule that enforces PHPUnit naming conventions

cs278:
#214

### DIFF
--- a/src/Rules/PHPUnit/ClassNamingRule.php
+++ b/src/Rules/PHPUnit/ClassNamingRule.php
@@ -10,7 +10,8 @@ use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
 use PHPUnit\Framework\TestCase;
 use function sprintf;
-use function str_ends_with;
+use function strlen;
+use function substr_compare;
 
 /**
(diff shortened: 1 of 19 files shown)


src/Rules/PHPUnit/ClassNamingRule.php
```diff
@@ -0,0 +1,111 @@
...
+			'Test',
+			'Concrete test class, \'%s\', should be named ending in \'%s\'.',
+		);
+
+		if (!$class->isFinal()) {
+			$errors[] = RuleErrorBuilder::message(sprintf(
+				'Concrete test class, \'%s\', should be declared final.',
+				$class->getName(),
+			))->identifier('phpunit.naming')->build();
+		}
```
staabm, line 71: I am not sure this "isFinal" thing is something which should be mixed with a "general naming rule".

whether tests should be final or not is a pretty opinionated decision IMO

ondrejmirtes:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it.

I've outlined why in this comment: https://github.com/phpstan/phpstan-phpunit/issues/214#issuecomment-2531190133

PR #217, opened 2024-12-13
Rules: Use more specific node-type

staabm:
as discussed in https://github.com/phpstan/phpstan-phpunit/pull/216#discussion_r1883733686


ondrejmirtes:
Thank you.

PR #219, opened 2024-12-19
Support `assertNotEquals` in AssertEqualsIsDiscouragedRule

staabm:
followup to https://github.com/phpstan/phpstan-phpunit/pull/216


staabm:
I think the build error is unrelated

ondrejmirtes:
Thank you.

PR #220, opened 2025-01-22
Fix error message for "assertNotEquals" usage referencing "assertSame" and "assertEquals"

PrinsFrank:
In #219 , "assertNotEquals" was added as discouraged. When there is a usage of "assertNotEquals" though, the error message still references "assertEquals" and "assertSame"

### DIFF
--- a/src/Rules/PHPUnit/AssertEqualsIsDiscouragedRule.php
+++ b/src/Rules/PHPUnit/AssertEqualsIsDiscouragedRule.php
@@ -11,6 +11,7 @@ use PHPStan\Type\GeneralizePrecision;
 use PHPStan\Type\TypeCombinator;
 use function count;
 use function in_array;
+use function sprintf;
 use function strtolower;
 
 /**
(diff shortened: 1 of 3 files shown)


src/Rules/PHPUnit/AssertEqualsIsDiscouragedRule.php
```diff
@@ -55,6 +55,14 @@ public function processNode(Node $node, Scope $scope): array
 			&& ($leftType->isSuperTypeOf($rightType)->yes())
 			&& ($rightType->isSuperTypeOf($leftType)->yes())
 		) {
+			if (strtolower($node->name->name) === 'assertnotequals') {
+				return [
+					RuleErrorBuilder::message(
+						'You should use assertNotSame() instead of assertNotEquals(), because both values are scalars of the same type',
```
ondrejmirtes: Would be nicer to just parametrize the existing return with `sprintf` instead of adding a whole new branch where most of the things are the same.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #221, opened 2025-03-12
fix build

staabm:


ondrejmirtes:
What does this fix?

staabm:
its about deprecation errors

ondrejmirtes:
In that case please install phpstan-deprecation-rules here.

ondrejmirtes:
Thank you.

PR #225, opened 2025-03-23
Add non regression test for #222

VincentLanglet:
With PHPUnit 12.0.9+, it gives
```
1) PHPStan\Rules\CallMethodsRuleTest::testBug222
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'
+'13: Call to method method() on an unknown class PHPUnit\Framework\MockObject\Builder\InvocationMocker.
+    💡 Learn more at https://phpstan.org/user-guide/discovering-symbols
+24: Call to method method() on an unknown class PHPUnit\Framework\MockObject\Builder\InvocationMocker.
+    💡 Learn more at https://phpstan.org/user-guide/discovering-symbols
 '
```


ondrejmirtes:
Thank you!

PR #235, opened 2025-09-02
Modernize code examples in README.md

derrabus:
This PR modernizes the code examples and the documentation on mock objects.

* Current PHPUnit releases use the namespaced interface `PHPUnit\Framework\MockObject\MockObject` for mocks. The non-namespaced `PHPUnit_Framework_MockObject_MockObject` is legacy and has been removed a long time ago.
* Current PHPUnit releases require the `setUp()` method in tests to have a `void` return type.
* PHP nowadays supports native intersection types which should be what we recommend to use in the README file. PHPDoc should only be a fallback if the projects needs to support legacy PHP runtimes.


ondrejmirtes:
Awesome, thank you!

PR #238, opened 2025-10-13
Implement DataProviderDataRule

### LINKED ISSUE #70: typecheck dataproviders

staabm:
Proof of concept

```php
<?php

use PHPUnit\Framework\Attributes\DataProvider;
use PHPUnit\Framework\TestCase;

class test extends TestCase
{

	#[DataProvider('aProvider')]
	public function testTrim(string $expectedResult, string $input): void
	{
	}

	public function aProvider(): array /** @phpstan-ignore missingType.iterableValue */
	{
		return [
			[
				'Hello World',
				" Hello World \n",
			],
			[
				'Hello World',
				123,
			],
			[
				'Hello World',
				false,
			],
		];
	}
}
```

leads to

```
 ------ ----------------------------------------------------------------------------- 
  Line   test.php                                                                     
 ------ ----------------------------------------------------------------------------- 
  21     Parameter #2 $input of method test::testTrim() expects string, int given.    
         🪪  argument.type                                                            
         at test.php:21                                                               
  25     Parameter #2 $input of method test::testTrim() expects string, false given.  
         🪪  argument.type                                                            
         at test.php:25                                                               
 ------ ----------------------------------------------------------------------------- 
```

Closes https://github.com/phpstan/phpstan-phpunit/issues/70

utilizes https://github.com/phpstan/phpstan-src/pull/4429 https://github.com/phpstan/phpstan-src/pull/4438

----

Scope
- same class dataprovider only for now

Todos
- [x] POC
- [x] Support `@test`
- [x] Support `#[Test]`
- [x] Support "test*" method name prefix
- [x] Support `@dataProvider`
- [x] Support `#[DataProvider]`
- [x] Support `static` data-provider
- [x] Support non-`static` data-provider
- [x] Support `return []` data-providers
- [x] Support `yield []` data-providers
- [x] Support `yield from []` data-providers
- [x] named arguments in data-provider
- [x] Bleeding Edge only
- [x] variadic parameters in test-methods
- [x] check number of arguments vs. parameters
- [x] Port `CompositionRule` into phpstan-src


tests/Rules/PHPUnit/data-provider-data.neon
```diff
@@ -0,0 +1,5 @@
+includes:
+	- ../../../extension.neon
+
+rules:
+	- PHPStan\Rules\Methods\CallMethodsRule
```
staabm: @ondrejmirtes I think we need something in RuleTestCase which allows me to add more then 1 rule, so the errors which will be generated by

```
$scope->invokeNodeCallback(new Node\Expr\MethodCall(
				$var,
				'testTrim',
				$args,
				['startLine' => $item->getStartLine()]
			));
```

will also show up in the test-suite. I tried manually registering an additional rule here, but this did not work

ondrejmirtes: Please implement CompositeRule here which can accept multiple rules in the constructor and will call them one-by-one. Once the PoC proves to work, we can add it to phpstan-src.

staabm:
Still work in progress. Will separate into more PRs after its mostly done

src/Rules/PHPUnit/DataProviderDataRule.php
```diff
@@ -0,0 +1,149 @@
...
+				$args = $this->arrayItemsToArgs($arrayExpr);
+				if ($args === null) {
+					continue;
+				}
+
+				if ($trimArgs && $maxNumberOfParameters !== $testMethod->getNumberOfParameters()) {
+					$args = array_slice($args, 0, min($testMethod->getNumberOfParameters(), $maxNumberOfParameters));
+				}
+
+				$var = new Node\Expr\New_(new Node\Name($classReflection->getName()));
```
ondrejmirtes: I'd rather do TypeExpr

src/Rules/PHPUnit/DataProviderDataRule.php
```diff
@@ -0,0 +1,149 @@
...
+
+	public function getNodeType(): string
+	{
+		return Node::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		if ($node instanceof Node\Stmt\Return_ || $node instanceof Node\Expr\YieldFrom) {
+			if (!$node->expr instanceof Node\Expr\Array_) {
```
ondrejmirtes: You should switch to typesystem logic asap, check whether it's a constant array...

src/Rules/PHPUnit/DataProviderDataRule.php
```diff
@@ -0,0 +1,149 @@
...
+	 */
+	private function arrayItemsToArgs(Node\Expr\Array_ $array): ?array
+	{
+		$args = [];
+
+		foreach ($array->items as $item) {
+			// XXX named args
+			$value = $item->value;
+
+			$arg = new Node\Arg($value);
```
ondrejmirtes: Again, TypeExpr will be your friend here, once you use types instead of AST 😊

ondrejmirtes (changes requested):

src/Rules/PHPUnit/DataProviderDataRule.php
staabm: adjusted.

src/Rules/PHPUnit/DataProviderDataRule.php
staabm: based on which type?

ondrejmirtes: New ObjectType with class reflection name

ondrejmirtes:
An idea: `yield from` does not require top-level array, just iterable. Think of the `yield from` statements in TypeInferenceTestCase child classes.

Maybe `return` doesn't need top-level array either, we should test that. (But it's useful to ask about constant arrays so that each item can be inspected separately.)

Maybe we should error when a non-array/non-iterable is returned, while respecting rule levels.

staabm:
> An idea: `yield from` does not require top-level array, just iterable. Think of the `yield from` statements in TypeInferenceTestCase child classes.

following example:

```php
class YieldFromExpr extends TestCase
{

	/** @dataProvider aProvider */
	public function testFoo(string $expectedResult, string $input): void
	{
	}

	public function aProvider(): iterable
	{
		yield from $this->moreData();
	}

	private function moreData(): iterable
	{
		yield [
			'Hello World',
			123,
		];
	}
}
```

do you mean in this case I should deep-inspect the constant returned arrays from `moreData` function?

staabm:
> Maybe `return` doesn't need top-level array either, we should test that. (But it's useful to ask about constant arrays so that each item can be inspected separately.)

if I understood correctly, this test proves that PHPUnit requires a wrapping top level array:

grafik

ondrejmirtes:
> mean in this case I should deep-inspect

No, just check if the exprType->getIterableValueType is a constant array (in case it's declared in PHPDoc).

ondrejmirtes:
> I understood correctly, this test proves

No, what I meant more is `return new ArrayIterator([[1], [2]])`

staabm:
> No, what I meant more is `return new ArrayIterator([[1], [2]])`

I think in this case we need to improve constant-array type inference before as we currently [don't get a constant array shape](https://phpstan.org/r/5ad703ce-b3e0-4f87-9f43-73465f880ddf)

staabm:
> No, just check if the exprType->getIterableValueType is a constant array (in case it's declared in PHPDoc).

test added in https://github.com/phpstan/phpstan-phpunit/pull/238/commits/7417fbc2889dcd1206b031cc5caf2cef50548830

staabm:
note sure what todo with all those compatibility errors

--

and especially this one I don't know what todo with it:

```
Parameter #1 $rules of class PHPStan\Testing\CompositeRule             
         constructor expects array<PHPStan\Rules\Rule<PhpParser\Node>>, array{  
         PHPStan\Rules\PHPUnit\DataProviderDataRule,                            
         PHPStan\Rules\Methods\CallMethodsRule} given.                          
         🪪  argument.type                                                       
         💡  Template type TNodeType on class PHPStan\Rules\Rule is not          
         covariant. Learn more: https://phpstan.org/blog/whats-up-with-templat  
         e-covariant                                                            
```

ondrejmirtes:
> I think in this case we need to improve constant-array type inference before as we currently

That's not necessary. Just make sure the rule will report something useful when we do `return $foo;` where `$foo` is `iterable<array{foo: string}>`.

src/Rules/PHPUnit/DataProviderDataRule.php
```diff
@@ -0,0 +1,179 @@
...
+		if ($scope->getFunction() === null) {
+			return [];
+		}
+
+		if ($scope->isInAnonymousFunction()) {
+			return [];
+		}
+
+		$method = $scope->getFunction();
+		if (!$method instanceof PhpMethodFromParserNodeReflection) {
```
ondrejmirtes: Just do `$method === null`

tests/Rules/PHPUnit/DataProviderDataRuleTest.php
```diff
@@ -0,0 +1,153 @@
+<?php declare(strict_types=1);
+
+namespace Rules\PHPUnit;
```
ondrejmirtes: Wrong namespace I think. That's why you get the BC promise errors.

ondrejmirtes (changes requested):
The `new CompositeRule` error will go away when this builds: https://github.com/phpstan/phpstan-src/commit/6e498e6d9a233a85393be2654cc829a552939b36

tests/Rules/PHPUnit/DataProviderDataRuleTest.php
staabm: nice catch! thanks

src/Rules/PHPUnit/DataProviderDataRule.php
```diff
@@ -0,0 +1,186 @@
...
+		if ($arraysTypes === []) {
+			return [];
+		}
+
+		$method = $scope->getFunction();
+		$classReflection = $scope->getClassReflection();
+		if (
+			$classReflection === null
+			|| !$classReflection->is(TestCase::class)
+			|| $classReflection->isAbstract()
```
ondrejmirtes: Why skip abstract tests? AFAIK the test method will run when the test is extended.

src/Rules/PHPUnit/DataProviderDataRule.php
```diff
@@ -0,0 +1,186 @@
...
+
+				if ($trimArgs && $maxNumberOfParameters !== $numberOfParameters) {
+					$args = array_slice($args, 0, min($numberOfParameters, $maxNumberOfParameters));
+				}
+
+				$scope->invokeNodeCallback(new Node\Expr\MethodCall(
+					new TypeExpr(new ObjectType($classReflection->getName())),
+					$testMethod->getName(),
+					$args,
+					['startLine' => $node->getStartLine()],
```
ondrejmirtes: It'd be great to find a better start line. If we have `return [[...], [...], ...]` that spans thousands of lines, it'd make this really usable. Not really necessary with `yield`.

There are cases where we won't have a precise line, like with `return $foo;`. I see two different approaches:

1) Let's write a special `if` when the node is `Return_` with `Array_`. And track the lines in `$arraysTypes`.
2) Keep the current typesystem approach but for each array, go back to the AST and try to find the right ArrayItem for each key-type. Possibly worse for performance.

src/Rules/PHPUnit/DataProviderDataRule.php
```diff
@@ -0,0 +1,186 @@
...
+		if ($node instanceof Node\Stmt\Return_ || $node instanceof Node\Expr\YieldFrom) {
+			if ($node->expr === null) {
+				return [];
+			}
+
+			$exprType = $scope->getType($node->expr);
+			$exprConstArrays = $exprType->getConstantArrays();
+			foreach ($exprConstArrays as $constArray) {
+				foreach ($constArray->getValueTypes() as $valueType) {
+					foreach ($valueType->getConstantArrays() as $constValueArray) {
```
ondrejmirtes: I think awkward things will happen when the count of `$exprType->getConstantArrays()` or `$valueType->getConstantArrays()` are > 1. Those are unions of arrays.

So you'll treat `array{1}|array{2}` as two separate test case invocations, but they're not really? It's more like a single invocation with `1|2`.

But I don't know what to do about arrays of different lengths.

ondrejmirtes (changes requested):

src/Rules/PHPUnit/DataProviderDataRule.php
staabm: I just added a few more const array tests. I did not yet experience something awkward.

do you have something specific in mind?

src/Rules/PHPUnit/DataProviderDataRule.php
staabm: re-added the ast based logic, when the Return_ or YieldFrom contains pure data

src/Rules/PHPUnit/DataProviderDataRule.php
staabm: I think the case you mentioned is already handled here: https://github.com/phpstan/phpstan-phpunit/pull/238/files#diff-84bda20337542c39c298550da2d95bfb9a60ef25f1e3a9ddc0e6840ce463664aR126

src/Rules/PHPUnit/TestMethodsHelper.php
```diff
@@ -0,0 +1,94 @@
...
+		FileTypeMapper $fileTypeMapper,
+		bool $phpunit10OrNewer
+	)
+	{
+		$this->fileTypeMapper = $fileTypeMapper;
+		$this->phpunit10OrNewer = $phpunit10OrNewer;
+	}
+
+	/**
+	 * @return array<ReflectionMethod>
```
ondrejmirtes, line 30: This is a weird type. I'd prefer our own ExtendedMethodReflection to be returned. You can get `ClassReflection::getNativeMethod()` once you make sure this is a test method from the current `$reflectionMethod` (to save time from instantiating all methods which is expensive).

ondrejmirtes: I noticed this because I was looking for how you're handling variadic methods and the `$testMethod->getNumberOfParameters()` method call seemed unfamiliar to me.

tests/Rules/PHPUnit/DataProviderDataRuleTest.php
```diff
@@ -0,0 +1,202 @@
...
+			self::getContainer()->getByType(CallMethodsRule::class) /** @phpstan-ignore phpstanApi.classConstant */
+		];
+
+		return new CompositeRule($rules);
+	}
+
+	public function testRule(): void
+	{
+		if (PHP_VERSION_ID < 80000) {
+			self::markTestSkipped();
```
ondrejmirtes: Why skip for PHP < 8.0?

ondrejmirtes (changes requested):
Also I pushed a bunch of failing tests 😊

src/Rules/PHPUnit/DataProviderDataRule.php
```diff
@@ -0,0 +1,237 @@
...
+				}
+			}
+		}
+
+		if (count($testsWithProvider) === 0) {
+			return [];
+		}
+
+		$maxNumberOfParameters = 0;
+		$trimArgs = count($testsWithProvider) > 1;
```
ondrejmirtes: I think in correct implementation, we don't need this condition at all. A foreach over methods that just iterates once should be enough to set everything correctly.

ondrejmirtes (changes requested):

VincentLanglet (commented):
What about code like
```
    #[DataProvider('aProvider')]
	public function testTrim(string $expectedResult, string $input): void
	{
	}

    /** @return array<array{string, string|int|false}> */
	public function aProvider(): array
	{
		return [
			[
				'Hello World',
				" Hello World \n",
			],
			[
				'Hello World',
				123,
			],
			[
				'Hello World',
				false,
			],
		];
	}
```
@staabm ?

I would expect:
- To not parse the implementation of the provider because I added a return type (which improves performance of the rule).
- `/** @return array<array{string, string|int|false}> */` to be reported because it's not compatible with the testTrim signature.

This way:
- I fix the phpdoc to `/** @return array<array{string, string}> */`
- And then, the `123` and `false` will be reported because they don't respect the `@return` type.

staabm:
I think with the new rule, you no longer need a return type on the provider (which is now redundant). having to define the signature at the test and another time at the provider feels like unnecessary work

IMO at best we would stop reporting "return type has no value type specified in iterable type array." for data-providers which can be analyzed by this PR.

VincentLanglet:
> I think with the new rule, you no longer need a return type on the provider (which is now redundant). having to define the signature at the test and another time at the provider fells like unnecessary work
> 
> IMO at best we would stop reporting "return type has no value type specified in iterable type array." for data-providers which can be analyzed by this PR.

I feel like it's a really opinionated decision, and might conflict with some other static analysis tool.
- I think psalm use the phpdoc from the dataprovider for his "similar" DataProviderDataRule.
- DataProvider might be extended/overriden and then I'll want to add phpdoc to restrict/extends the return type.
- Team could have a 100% typed policy (like we do)

Couldn't the DataProviderDataRule either
- Skip provider with phpdoc (and another rule will do the check)
- Or rely only on the phpdoc

This rule will 

Cause some code like
```
    #[DataProvider('aProvider')]
	public function testTrim(string $expectedResult, string $input): void
	{
	}

    /** @return array<array{string, string|int}> */
	public function aProvider(): array
	{
		return [
			[
				'Hello World',
				" Hello World \n",
			],
		];
	}
```
Is not reported by the rule currently, but technically wrong (and reported by Psalm)

ondrejmirtes:
@VincentLanglet I disagree with you but I'm on a phone and might write a proper response latee.

src/Rules/PHPUnit/DataProviderDataRule.php
```diff
@@ -0,0 +1,237 @@
...
+		}
+
+		$method = $scope->getFunction();
+		$classReflection = $scope->getClassReflection();
+		if (
+			$classReflection === null
+			|| !$classReflection->is(TestCase::class)
+		) {
+			return [];
+		}
```
VincentLanglet, line 70: Shouldn't this condition be done before buildArrayTypesFromNode to be faster ?

We might also prefer to check
```
if (count($testsWithProvider) === 0) {
     return [];
}
```
before the `buildArrayTypesFromNode` call, it will avoid to inspect methods which are not dataprovider and improves performances.

src/Rules/PHPUnit/DataProviderDataRule.php
```diff
@@ -0,0 +1,237 @@
...
+		if ($scope->isInAnonymousFunction()) {
+			return [];
+		}
+
+		$arraysTypes = $this->buildArrayTypesFromNode($node, $scope);
+		if ($arraysTypes === []) {
+			return [];
+		}
+
+		$method = $scope->getFunction();
```
VincentLanglet, line 63: Since dataprovider need to be public, what about an early return if the method is protected/private ?

staabm: depends on the phpunit version range we want to support. in phpunit 9.x dataproviders are allowed to be `protected` or even `private` and also non-static

ondrejmirtes: Also the method is still registered as a data provider. I'd still typecheck the items.

Because it doesn't feel great to fix 1 error and have 100 more appear. It's better to have 101 reported from the beginning.

src/Rules/PHPUnit/DataProviderDataRule.php
staabm: I am not sure whether AST traversing + type retrieval is faster or slower then runtime reflection.
do you have a test at hand which to prove your thesis about what is faster?

VincentLanglet: > I am not sure whether AST traversing + type retrieval is faster or slower then runtime reflection.
do you have a cast at hand which to prove your thesis about what is faster?

Not at all, it might be a wrong intuition.
I thought type manipulation was something slow, with possible union/intersection etc.

src/Rules/PHPUnit/DataProviderDataRule.php
```diff
@@ -0,0 +1,237 @@
...
+			$classReflection === null
+			|| !$classReflection->is(TestCase::class)
+		) {
+			return [];
+		}
+
+		$testsWithProvider = [];
+		$testMethods = $this->testMethodsHelper->getTestMethods($classReflection, $scope);
+		foreach ($testMethods as $testMethod) {
+			foreach ($this->dataProviderHelper->getDataProviderMethods($scope, $testMethod, $classReflection) as [, $providerMethodName]) {
```
VincentLanglet, line 75: Do you think it would be worth having a cache on `getDataProviderMethods` (and maybe getTestMethods ?) ?

Cause we will compute this for every possible return of a TestCase class, and the result will always be the same

staabm: I am fine/willing to add a cache in case we find a case where it turns into a measurable improvement

src/Rules/PHPUnit/TestMethodsHelper.php
staabm: from `ExtendedMethodReflection` I cannot `getStartLine`, which I need in `DataProviderHelper`

tests/Rules/PHPUnit/DataProviderDataRuleTest.php
staabm: separated the named argument tests into a new file

src/Rules/PHPUnit/TestMethodsHelper.php
ondrejmirtes: You mean there isn't `getStartLine` on AttributeReflection? I don't see `getStartLine` being called on `ReflectionMethod`.

ondrejmirtes:
@VincentLanglet To reply to your earlier comment about this code:

```php
    #[DataProvider('aProvider')]
	public function testTrim(string $expectedResult, string $input): void
	{
	}

    /** @return array<array{string, string|int}> */
	public function aProvider(): array
	{
		return [
			[
				'Hello World',
				" Hello World \n",
			],
		];
	}
```

Currently this code will be reported with return.nestedUnusedType (https://phpstan.org/r/3b73fe7f-361e-4f8e-a752-0cf8827db30e). Provided it's in a final class or you turn on `checkTooWideReturnTypesInProtectedAndPublicMethods`.

src/Rules/PHPUnit/TestMethodsHelper.php
staabm: see https://github.com/phpstan/phpstan-phpunit/blob/39950c70b6a4faa36567b83b8c4b32f889c68e03/src/Rules/PHPUnit/DataProviderHelper.php#L292

staabm:
> Also I pushed a bunch of failing tests 😊

fixed, thank you 🙏

ondrejmirtes:
I played as Infection for a bit and found some code that could be removed from the algorithm 😊 I still think there's something fishy about it but I didn't prove it yet. In my gut I think we should be interesting in the min number of parameters when deciding how to trim, but maybe I'm wrong, or maybe I will find another failing test :) We'll see. Thank you so far!

staabm:
> In my gut I think we should be interesting in the min number of parameters when deciding how to trim

thats what you deleted with https://github.com/phpstan/phpstan-phpunit/pull/238/commits/f083e6319a52692b7dfa5ddcc97c7b8f74c49cbf ...? :)

staabm:
> I played as Infection for a bit

understood. finished the infection package, so we can setup it :)

https://github.com/phpstan/build-infection/pull/1

ondrejmirtes:
> thats what you deleted

I deleted it but the code didn't do anything, no tests failed. I'll start looking into this again.

ondrejmirtes:
Awesome, thank you!

PR #239, opened 2025-10-14
Preparational refactoring for DataProviderDataRule

staabm:


ondrejmirtes (changes requested):
Maybe I'm too slow today - but why do we need PHPUnit version detection for data providers? There are versions that support both `@dataProvider` PHPDoc tag and `#[DataProvider]` attribute. We should be looking at both.

staabm:
> Maybe I'm too slow today - but why do we need PHPUnit version detection for data providers? There are versions that support both `@dataProvider` PHPDoc tag and `#[DataProvider]` attribute. We should be looking at both.

in this PR I am mostly refactoring pre-existing logic.

while I agree from a functional point of view, its not necessary, but I think we could improve performance with this knowledge (what is not implemented right now) - because there are PHPUnit versions which support only annotations, some support both and newer ones only support Attributes (no longer annotations). so we could prevent unnecessary phpdoc parsing work on newer PHPUnit

ondrejmirtes:
@staabm I don't think we should use this knowledge to improve performance (it'd be negligible). I think we should use it to report "`@dataProvider` is deprecated on PHPUnit 11.x" (when `%deprecationRulesInstalled%` is true) and on PHPUnit 12.x to report that it's no longer supported 😊

ondrejmirtes:
Thank you.

Issue #231, opened 2025-05-23
assertStringStartsWith after assertNotEmpty fails


gisostallenberg:
assertStringStartsWith after assertNotEmpty fails with error "Parameter #1 $prefix of method PHPUnit\Framework\Assert::assertStringStartsWith() expects non-empty-string', while the string cannot be empty due to the `assertNotEmpty` call.

```php
<?php

use PHPUnit\Framework\TestCase;

class FooTest extends TestCase {
    public function testFoo(): void
    {
        $string = ([
            '',
            'bar',
        ])[rand(0, 1)];
        $this->assertNotEmpty($string);
        $this->assertStringStartsWith($string, 'baz');
    }
}
```

ondrejmirtes:
Can't reproduce the problem: https://phpstan.org/r/91888ed1-d67f-4749-9d9c-a617da10722b

Please show the output of `vendor/bin/phpstan diagnose`.

ondrejmirtes:
I get the problem now, phpstan-phpunit has a special type-specifying extension that doesn't work for strings: https://github.com/phpstan/phpstan-phpunit/blob/d35895e0388e9aa822af00e1dfd4f8086fd8938d/src/Type/PHPUnit/Assert/AssertTypeSpecifyingExtensionHelper.php#L168-L177

staabm:
I was not able to reproduce the issue, but I think https://github.com/phpstan/phpstan-src/pull/4473 will have fixed it in case its still relevant

staabm:
@gisostallenberg please double check whether this issue still reproduces on the latest release

PR #241, opened 2025-10-24
Added assertArrayHasKey() expectations

### LINKED ISSUE #100: AssertArrayHasKey seems to be not supported
Hi @ondrejmirtes, I just got an issue with my tests saying that an offset does not exist in an array.
I tried to add a `assertArrayHasKey` check, without success.

I think this assert is not handle by this plugin, am I wrong ?

I'll try to do a PR in order to add the support, I think I'll be able to copy the way `isArray` works.
https://github.com/phpstan/phpstan-phpunit/blob/master/src/Type/PHPUnit/Assert/AssertTypeSpecifyingExtensionHelper.php#L158

staabm:
will be fixed when https://github.com/phpstan/phpstan-src/pull/4473 is merged

closes https://github.com/phpstan/phpstan-phpunit/issues/100

### DIFF
--- a/tests/Type/PHPUnit/data/assert-function.php
+++ b/tests/Type/PHPUnit/data/assert-function.php
@@ -11,7 +11,6 @@ use function PHPUnit\Framework\assertCount;
 use function PHPUnit\Framework\assertNotCount;
 use function PHPUnit\Framework\assertEmpty;
 use function PHPUnit\Framework\assertInstanceOf;
-use function PHPUnit\Framework\assertObjectHasAttribute;
 
 class Foo
 {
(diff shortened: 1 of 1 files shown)


staabm:
memo to me: might also fix https://github.com/phpstan/phpstan-phpunit/issues/231 - need to verify after phpstan-src PR merged

staabm:
yay, fixed

ondrejmirtes:
Can you add a test for #231 and make sure it fails with current stable PHPStan release?

staabm:
I tried reproducing the issue but wasn't able to get the reported problem

ondrejmirtes:
So we don't know whether and when it was fixed

ondrejmirtes:
Thank you!

PR #243, opened 2025-10-26
Setup mutation testing

staabm:
analog
- https://github.com/phpstan/phpstan-doctrine/pull/687
- https://github.com/phpstan/phpstan-doctrine/pull/686
- https://github.com/phpstan/phpstan-doctrine/pull/699

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -236,15 +236,10 @@ jobs:
       - name: "Checkout"
         uses: actions/checkout@v5
 
-      - name: "Install PHP"
-        uses: "shivammathur/setup-php@v2"
+      - uses: ./build-infection/.github/actions/setup-php
         with:
-          coverage: "pcov"
           php-version: "${{ matrix.php-version }}"
-          tools: pecl, infection:0.31.7
           extensions: mbstring
-          ini-file: development
-          ini-values: memory_limit=-1
 
       - name: "Install dependencies"
         run: "composer install --no-interaction --no-progress"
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -213,3 +217,72 @@ jobs:
...
+          - "8.2"
+          - "8.3"
+          - "8.4"
+        operating-system: [ubuntu-latest]
+
+    steps:
+      - name: "Checkout"
+        uses: actions/checkout@v5
+
+      - name: "Install PHP"
```
ondrejmirtes: Please use the composite action, same way it's used here (the commit first moves "Checkout build-infection" right after "Checkout", then we can use a local path to the composite action): https://github.com/phpstan/phpstan-doctrine/pull/704

.github/workflows/build.yml
```diff
@@ -213,3 +217,72 @@ jobs:
...
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "pcov"
+          php-version: "${{ matrix.php-version }}"
+          tools: pecl, infection:0.31.7
```
ondrejmirtes: Why pecl? It's not used anywhere else.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

ondrejmirtes:
Also failing https://github.com/phpstan/phpstan-phpunit/actions/runs/18823300049

PR #245, opened 2025-10-27
move analyse-paths into phpstan.neon

staabm:
should fix

```
In InitialStaticAnalysisRunFailed.php line 29:
                                                                               
  Project static analysis must be in a passing state before running Infection  
  .                                                                            
  PHPStan reported an exit code of 1.                                          
  Refer to the PHPStan's output below:                                         
  STDERR:                                                                      
  At least one path must be specified to analyse.           
  ```

in https://github.com/phpstan/build-infection/pull/15

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,8 +26,8 @@ cs-fix:
 
 .PHONY: phpstan
 phpstan:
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Thank you!

PR #246, opened 2025-10-27
Ignore missingType.iterableValue for data-providers

staabm:
As we are now [validating data-providers by its return/yield statements](https://github.com/phpstan/phpstan-phpunit/pull/238), I think we should no longer throw a 

> Method DataProviderIterableValueTest\Foo::dataProvider() return type has no value type specified in iterable type iterable.

error onto the PHPStan user

---

requires https://github.com/phpstan/phpstan-src/pull/4488


src/Type/PHPUnit/DataProviderReturnTypeIgnoreExtension.php
```diff
@@ -0,0 +1,53 @@
...
+		DataProviderHelper $dataProviderHelper
+	)
+	{
+		$this->testMethodsHelper = $testMethodsHelper;
+		$this->dataProviderHelper = $dataProviderHelper;
+	}
+
+	public function shouldIgnore(Error $error, Node $node, Scope $scope): bool
+	{
+		if (! $node instanceof InClassMethodNode) { // @phpstan-ignore phpstanApi.instanceofAssumption
```
ondrejmirtes: This assumes an implementation detail that missingType.iterableValue is reported by a rule that hooks onto InClassMethodNode.

You can remove this instanceof and just ask for `$scope->isInClass()`, `$scope->getFunction()` etc.

ondrejmirtes (changes requested):

ondrejmirtes:
BTW I love this, it's genius :)

src/Type/PHPUnit/DataProviderReturnTypeIgnoreExtension.php
staabm: got this code from the docs :).

doc-fix in https://github.com/phpstan/phpstan/pull/13726

thank you.

ondrejmirtes:
Thank you!

VincentLanglet:
@staabm @ondrejmirtes Is there an easy way to (?)
- take advantage of the DataProviderDataRule
- without enabling the DataProviderReturnTypeIgnoreExtension

Would it be ok to have such option ?

Maybe like
```
checkDataProviderData: %featureToggles.bleedingEdge%
ignoreDataProviderReturnType: %featureToggles.bleedingEdge%

PHPStan\Type\PHPUnit\DataProviderReturnTypeIgnoreExtension:
		phpstan.ignoreErrorExtension: %phpunit.ignoreDataProviderReturnType%
```
or
```
checkDataProviderData: %featureToggles.bleedingEdge%
checkDataProviderReturnType: false

PHPStan\Type\PHPUnit\DataProviderReturnTypeIgnoreExtension:
		phpstan.ignoreErrorExtension: [%phpunit.ignoreDataProviderReturnType%, not(%phpunit.checkDataProviderReturnType%)]
```
if such syntax exists...

In our codebase, we appreciate the fact PHPStan enforce the detailed phpdoc of iterable (for documentation purpose), and even if the return/yield statement are now checked, we still find some interest in the required phpdoc.

ondrejmirtes:
So I realize I don't want to have a long-term option for turning off DataProviderDataRule. I only want it to be conditional now in 2.x based on bleedingEdge.

As for  the data provider return types, I can imagine there to be an option named something like `reportMissingDataProviderReturnType` that would default to false. You could set it to true to disable the ignore errors extension.

Feel free to send a PR.

VincentLanglet:
> Feel free to send a PR.

Like this https://github.com/phpstan/phpstan-phpunit/pull/253 ?

PR #247, opened 2025-10-27
Assert*Rules: Do cheap checks first

staabm:
shuffling the initial if-checks so the cheapest ones get called first.

`AssertRuleHelper::isMethodOrStaticCallOnAssert` internally uses `new ObjectType('PHPUnit\Framework\Assert')` which triggers reflection logic and file parsing

see

grafik

---

refs https://github.com/phpstan/build-infection/issues/16

### DIFF
--- a/src/Rules/PHPUnit/AssertEqualsIsDiscouragedRule.php
+++ b/src/Rules/PHPUnit/AssertEqualsIsDiscouragedRule.php
@@ -27,6 +27,9 @@ class AssertEqualsIsDiscouragedRule implements Rule
 
 	public function processNode(Node $node, Scope $scope): array
 	{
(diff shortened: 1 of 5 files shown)


src/Rules/PHPUnit/AssertEqualsIsDiscouragedRule.php
```diff
@@ -27,24 +27,27 @@ public function getNodeType(): string
 
 	public function processNode(Node $node, Scope $scope): array
 	{
-		if (!AssertRuleHelper::isMethodOrStaticCallOnAssert($node, $scope)) {
```
staabm, line 30: before this PR, any CallLike triggered autoloading of the PHPUnit classes via reflection (no matter whether it actually contained the `assert` prefix in the method name or no matter whether the CallLike was related to PHPUnit at all)

ondrejmirtes:
Thank you!

PR #248, opened 2025-10-28
Refactor PHPUnitVersionDetector to ease different major version checks

staabm:
required for https://github.com/phpstan/phpstan-src/pull/4490#issuecomment-3455197803


ondrejmirtes:
I'd like a PHPUnitVersion class that works the same way PhpVersion works in PHPStan. Methods asking about specific features.

staabm:
here you are

ondrejmirtes:
The majorVersion is nullable when we can't detect it? I wonder if we should make the user of PHPUnitVersion explicitly handle it, because otherwise we're basically lying. When we can't detect it, the thing might be supported.

staabm:
Ok. Adding trinary then

staabm:
I just had my first mutation testing experience on a phpstan/* repo :)

ondrejmirtes:
Perfect, thank you!

ondrejmirtes:
> I just had my first mutation testing experience on a phpstan/* repo 😊

Let's record this here: https://github.com/phpstan/phpstan-phpunit/actions/runs/18886595907/job/53903464005

The log will disappear after some time so here's a screenshot:

Screenshot 2025-10-29 at 9 53 02

This is what I had in mind the whole time, I'm glad it pays off 👏

PR #250, opened 2025-10-29
PHPUnit 9.x/10.x does not at all support named arguments from data-providers

staabm:
prevents phpstan-src errors we currently see on PHP 7.x:

```
------ ------------------------------------------------------------------ 
  Line   tests/PHPStan/Command/ErrorFormatter/TableErrorFormatterTest.php  
 ------ ------------------------------------------------------------------ 
  44     Named arguments are supported only on PHP 8.0 and later.          
  57     Named arguments are supported only on PHP 8.0 and later.          
  76     Named arguments are supported only on PHP 8.0 and later.          
  95     Named arguments are supported only on PHP 8.0 and later.          
  124    Named arguments are supported only on PHP 8.0 and later.          
  144    Named arguments are supported only on PHP 8.0 and later.          
  180    Named arguments are supported only on PHP 8.0 and later.          
  199    Named arguments are supported only on PHP 8.0 and later.          
  220    Named arguments are supported only on PHP 8.0 and later.          
  241    Named arguments are supported only on PHP 8.0 and later.          
 ------ ------------------------------------------------------------------ 
 ```

refs https://github.com/phpstan/phpstan-src/pull/4490#issuecomment-3455197803

### DIFF
--- a/tests/Rules/PHPUnit/DataProviderDataRuleTest.php
+++ b/tests/Rules/PHPUnit/DataProviderDataRuleTest.php
@@ -308,11 +308,14 @@ class DataProviderDataRuleTest extends RuleTestCase
 
 	static public function provideNamedArgumentPHPUnitVersions(): iterable
 	{
-		return [
-			[null],
-			[10],
-			[11],
-		];
+		yield [null]; // unknown phpunit version
+
+		if (PHP_VERSION_ID >= 80100) {
+			yield [10]; // PHPUnit 10.x requires PHP 8.1+
+		}
+		if (PHP_VERSION_ID >= 80200) {
+			yield [11]; // PHPUnit 11.x requires PHP 8.2+
+		}
 	}
 
 	/**
@@ -324,10 +327,6 @@ class DataProviderDataRuleTest extends RuleTestCase
 		$this->phpunitVersion = $phpunitVersion;
 
 		if ($phpunitVersion >= 11) {
-			if (PHP_VERSION_ID < 80000) {
-				self::markTestSkipped('PHPUnit11 requires PHP 8.0+');
-			}
-
 			$errors = [];
 		} else {
 			$errors = [
--- a/tests/Rules/PHPUnit/DataProviderDataRuleTest.php
+++ b/tests/Rules/PHPUnit/DataProviderDataRuleTest.php
@@ -179,9 +179,9 @@ class DataProviderDataRuleTest extends RuleTestCase
 
 
 	/**
-	 * @dataProvider provideNamedArgumentVersions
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: tests/Rules/PHPUnit/DataProviderDataRuleTest.php [hunk 1]
```diff
@@ -176,30 +179,26 @@ public function testRule(): void
...
+				'Parameter #1 $expectedResult of method DataProviderDataTestPhp8\NamedArgsInProvider::testFoo() expects string, false given.',
 				44
 			],
 			[
-				'Unknown parameter $wrong in call to method DataProviderDataTestPhp8\TestWrongOffsetNameArrayShapeIterable::testBar().',
+				'Parameter #1 $si of method DataProviderDataTestPhp8\TestWrongOffsetNameArrayShapeIterable::testBar() expects int, string given.',
 				58
 			],
 			[
-				'Missing parameter $si (int) in call to method DataProviderDataTestPhp8\TestWrongOffsetNameArrayShapeIterable::testBar().',
```
ondrejmirtes, line 198: Some tests for "missing parameter"/"unknown parameter" would be nice though.

ondrejmirtes (changes requested):

tests/Rules/PHPUnit/DataProviderDataRuleTest.php
```diff
@@ -274,6 +306,44 @@ public function testTrimmingArgs(): void
 		]);
 	}
 
+	static public function provideNamedArgumentPHPUnitVersions(): iterable
+	{
+		yield [null]; // unknown phpunit version
+
+		if (PHP_VERSION_ID >= 80100) {
+			yield [10]; // PHPUnit 10.x requires PHP 8.1+
```
staabm, line 314: hmm.. maybe we should even reflect this in PHPUnitVersion.

that way we could turn a `maybe` into a `no` depending on the php runtime version?

ondrejmirtes: I thought about it but probably no.

The only PHPUnit versions that support named arguments (in the sense they do not do `array_values()` on `...$args` also only support PHP 8+.

The only wrong thing that can happen I think is when we're not successfully detecting PHPUnit version, and the user is running PHPUnit 11.5+ - we will produce `Arg` without names so possibly the result is going to be wrong.

But I don't think a condition about PHP version belongs to the logic. Because it's unrelated - you can run PHPUnit 9.x on PHP 8 as well...

ondrejmirtes:
Thank you! Let's update the extension in the phpstan-src PR to see if it's going to be green :)

Issue #249, opened 2025-10-29
turn `AssertSame*` rules autofixable


staabm:
[`AssertEqualsIsDiscouragedRule` got autofixable recently](https://github.com/phpstan/phpstan-phpunit/commit/9a9b161baee88a5f5c58d816943cff354ff233dc).

we should do the same with the remaining `AssertSame*` rules

PR #252, opened 2025-10-30
Make AssertSameBooleanExpectedRule auto-fixable

staabm:
refs https://github.com/phpstan/phpstan-phpunit/issues/249

### DIFF
--- a/src/Rules/PHPUnit/AssertEqualsIsDiscouragedRule.php
+++ b/src/Rules/PHPUnit/AssertEqualsIsDiscouragedRule.php
@@ -73,13 +73,7 @@ class AssertEqualsIsDiscouragedRule implements Rule
 					),
 				)->identifier('phpunit.assertEquals')
 					->fixNode($node, static function (CallLike $node) use ($correctName) {
-						if ($node instanceof Node\Expr\MethodCall) {
-							$node->name = new Node\Identifier($correctName);
-						}
-
-						if ($node instanceof Node\Expr\StaticCall) {
-							$node->name = new Node\Identifier($correctName);
-						}
+						$node->name = new Node\Identifier($correctName);
 
 						return $node;
 					})
--- a/src/Rules/PHPUnit/AssertSameBooleanExpectedRule.php
+++ b/src/Rules/PHPUnit/AssertSameBooleanExpectedRule.php
@@ -47,13 +47,35 @@ class AssertSameBooleanExpectedRule implements Rule
 
 		if ($expectedArgumentValue->name->toLowerString() === 'true') {
 			return [
-				RuleErrorBuilder::message('You should use assertTrue() instead of assertSame() when expecting "true"')->identifier('phpunit.assertTrue')->build(),
+				RuleErrorBuilder::message('You should use assertTrue() instead of assertSame() when expecting "true"')
+					->identifier('phpunit.assertTrue')
+					->fixNode($node, static function (CallLike $node) {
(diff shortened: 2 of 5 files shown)


src/Rules/PHPUnit/AssertEqualsIsDiscouragedRule.php
```diff
@@ -73,13 +73,7 @@ public function processNode(Node $node, Scope $scope): array
 					),
 				)->identifier('phpunit.assertEquals')
 					->fixNode($node, static function (CallLike $node) use ($correctName) {
-						if ($node instanceof Node\Expr\MethodCall) {
```
staabm, line 76: since a recent change we have 
```php
if (!$node instanceof Node\Expr\MethodCall && ! $node instanceof Node\Expr\StaticCall) {
			return [];
}
```

at the very top of this rule, therefore we don't need to narrow here again anymore.

src/Rules/PHPUnit/AssertSameBooleanExpectedRule.php
```diff
@@ -47,13 +47,35 @@ public function processNode(Node $node, Scope $scope): array
...
 			return [
-				RuleErrorBuilder::message('You should use assertTrue() instead of assertSame() when expecting "true"')->identifier('phpunit.assertTrue')->build(),
+				RuleErrorBuilder::message('You should use assertTrue() instead of assertSame() when expecting "true"')
+					->identifier('phpunit.assertTrue')
+					->fixNode($node, static function (CallLike $node) {
+						$node->name = new Node\Identifier('assertTrue');
+
+						$args = $node->getArgs();
+						unset($args[0]);
+						$node->args = $args;
```
ondrejmirtes: Pretty sure you're breaking the args list contract here 😊

src/Rules/PHPUnit/AssertSameBooleanExpectedRule.php
```diff
@@ -47,13 +47,35 @@ public function processNode(Node $node, Scope $scope): array
...
 		if ($expectedArgumentValue->name->toLowerString() === 'false') {
 			return [
-				RuleErrorBuilder::message('You should use assertFalse() instead of assertSame() when expecting "false"')->identifier('phpunit.assertFalse')->build(),
+				RuleErrorBuilder::message('You should use assertFalse() instead of assertSame() when expecting "false"')
+					->identifier('phpunit.assertFalse')
+					->fixNode($node, static function (CallLike $node) {
+						$node->name = new Node\Identifier('assertFalse');
+
+						$args = $node->getArgs();
+						unset($args[0]);
```
ondrejmirtes: dtto

ondrejmirtes (changes requested):

src/Rules/PHPUnit/AssertSameBooleanExpectedRule.php
staabm: good point.

what do you think about moving the new PHPParser stubs into the regular phpstan-src stubs/ directory and load them conditional via a new phpstan-src config parameter, so we can enable the more precise types in 1st party extension repos.

ondrejmirtes: @staabm I'm not sure about it. I don't feel like adding a configuration option and documenting it is justified just as a temporary measure before your PR on PHP-Parser is merged.

ondrejmirtes:
Thank you!

PR #254, opened 2025-11-02
Use TypeSystem in AssertSameBooleanExpectedRule

staabm:
I think it makes sense that we get the same error when working with constants and known value.

https://github.com/phpstan/phpstan-phpunit/blob/450bfc08e2e5d00b214110de14168d4ebd210cfb/tests/Rules/PHPUnit/data/assert-same-boolean-expected.php#L29-L33

### DIFF
--- a/src/Rules/PHPUnit/AssertSameBooleanExpectedRule.php
+++ b/src/Rules/PHPUnit/AssertSameBooleanExpectedRule.php
@@ -40,18 +40,19 @@ class AssertSameBooleanExpectedRule implements Rule
 		if (!($expectedArgumentValue instanceof ConstFetch)) {
 			return [];
 		}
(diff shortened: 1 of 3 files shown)


staabm:
another case in which infection made me add new tests:

grafik

ondrejmirtes:
I disagree with this. I feel like that when someones does:

```php
$this->assertSame(Foo::BAR, $expected);
```

It's there on purpose and it means "expected should be the same as Foo::BAR". The value of Foo::BAR might be true right now but it's there so that the value can change.

Changing the code to:

```php
$this->assertTrue($expected);
```

would lose information.

staabm:
ok. reduced it to the new tests

staabm:
remaining tests are here to kill mutants

ondrejmirtes:
Thank you!

PR #257, opened 2025-11-11
Fix AssertSameWithCountRule for recursive count()

staabm:
inspired by https://github.com/phpstan/phpstan-src/pull/4515


src/Rules/PHPUnit/AssertSameWithCountRule.php
```diff
@@ -42,36 +46,67 @@ public function processNode(Node $node, Scope $scope): array
 		}
 
 		$right = $node->getArgs()[1]->value;
-		if (
-			$right instanceof Node\Expr\FuncCall
```
staabm, line 46: extracted conditions logic into new methods, to ease implementing auto-fixable in a followup

ondrejmirtes:
Thank you!

PR #258, opened 2025-11-11
Fix "Unexpected input(s) 'extensions', valid inputs are ['php-version', 'php-extensions', 'build-infection-path']"

staabm:


ondrejmirtes:
Thank you!

PR #259, opened 2025-11-11
Make AssertSameWithCountRule auto-fixable

### LINKED ISSUE #249: turn `AssertSame*` rules autofixable

staabm:
closes https://github.com/phpstan/phpstan-phpunit/issues/249


staabm:
@ondrejmirtes here another low hanging fruit :-)

ondrejmirtes (changes requested):
A fix and a test for tricky situations with named arguments would be nice here 😊

PR #268, opened 2025-12-05
Implement AttributeRequiresPhpVersionRule

staabm:
using `#[RequiresPhp('8.3')]` without a version operator like `>=`, `^` or similar is deprecated as of PHPUnit 12.4 and was removed in PHPUnit 13.0

refs https://github.com/sebastianbergmann/phpunit/issues/6355

### DIFF
--- a/tests/Rules/PHPUnit/data/requires-php-version.php
+++ b/tests/Rules/PHPUnit/data/requires-php-version.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace RequiresPhpVersion;
+
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\Test;
+use PHPUnit\Framework\TestCase;
+use PHPUnit\Framework\Attributes\RequiresPhp;
+
+class DeprecatedVersionFormat extends TestCase
+{
+	#[RequiresPhp('8.0')]
+	public function testDeprecatedFormat(): void {
+
+	}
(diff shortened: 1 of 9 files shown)


tests/Rules/PHPUnit/data/requires-php-version.php
```diff
@@ -0,0 +1,24 @@
...
+namespace RequiresPhpVersion;
+
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\Test;
+use PHPUnit\Framework\TestCase;
+use PHPUnit\Framework\Attributes\RequiresPhp;
+
+class DeprecatedVersionFormat extends TestCase
+{
+	#[RequiresPhp('8.0')]
```
staabm, line 12: in a followup I want to emit errors when the version contstraint used within `#[RequiresPhp('8.0')]` does not fit into the php version used for analysis (detecting tests which require php-versions no longer supported in the project)

ondrejmirtes:
Yeah, this is really needed, I made this mistake a few times myself.

ondrejmirtes:
Thank you!

tests/Rules/PHPUnit/data/requires-php-version.php
staabm: implemented in https://github.com/phpstan/phpstan-phpunit/pull/269

PR #269, opened 2025-12-05
Sanity check `#[RequiresPhp]` value and range

staabm:
- when `#[RequiresPhp]` is lower then the phpstan analysis version, reports a test requirement as beeing always false
- when `#[RequiresPhp]` contains a version which is not parse-able, reports a error

### DIFF
--- a/src/Rules/PHPUnit/AttributeRequiresPhpVersionRule.php
+++ b/src/Rules/PHPUnit/AttributeRequiresPhpVersionRule.php
@@ -2,11 +2,8 @@
 
 namespace PHPStan\Rules\PHPUnit;
 
-use Composer\Semver\Constraint\ConstraintInterface;
-use Composer\Semver\VersionParser;
 use PharIo\Version\UnsupportedVersionConstraintException;
 use PharIo\Version\Version;
-use PharIo\Version\VersionConstraint;
 use PharIo\Version\VersionConstraintParser;
 use PhpParser\Node;
 use PHPStan\Analyser\Scope;
@@ -15,20 +12,20 @@ use PHPStan\Php\PhpVersion;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
 use PHPUnit\Framework\TestCase;
-use UnexpectedValueException;
 use function count;
 use function is_numeric;
 use function method_exists;
 use function preg_match;
 use function sprintf;
+use function version_compare;
 
 /**
  * @implements Rule<InClassMethodNode>
  */
 class AttributeRequiresPhpVersionRule implements Rule
 {
-	private const VERSION_COMPARISON = "/(?P<operator>!=|<|<=|<>|=|==|>|>=)?\s*(?P<version>[\d\.-]+(dev|(RC|alpha|beta)[\d\.])?)[ \t]*\r?$/m";
 
+	private const VERSION_COMPARISON = "/(?P<operator>!=|<|<=|<>|=|==|>|>=)?\s*(?P<version>[\d\.-]+(dev|(RC|alpha|beta)[\d\.])?)[ \t]*\r?$/m";
 
 	private Version $phpstanPhpVersion;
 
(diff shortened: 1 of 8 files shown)


src/Rules/PHPUnit/AttributeRequiresPhpVersionRule.php
```diff
@@ -71,6 +82,29 @@ public function processNode(Node $node, Scope $scope): array
 			if (
 				!is_numeric($args[0])
 			) {
+
+				try {
+					$testPhpVersionConstraint = $parser->parseConstraints($args[0]);
+				} catch (UnexpectedValueException $e) {
+					$errors[] = RuleErrorBuilder::message(
+						sprintf($e->getMessage()),
```
staabm: not sure we want to 1:1 print the message from composer/semver as a phpstan rule

src/Rules/PHPUnit/AttributeRequiresPhpVersionRule.php
```diff
@@ -31,12 +37,16 @@ class AttributeRequiresPhpVersionRule implements Rule
...
+		bool $deprecationRulesInstalled,
+		PhpVersion $phpVersion
 	)
 	{
 		$this->PHPUnitVersion = $PHPUnitVersion;
 		$this->testMethodsHelper = $testMethodsHelper;
 		$this->deprecationRulesInstalled = $deprecationRulesInstalled;
+
+		$parser = new VersionParser();
+		$this->phpstanVersionConstraint = $parser->parseConstraints($phpVersion->getVersionString());
```
staabm: maybe this should compare against the composer.json min version instead of the phpstan.neon version?

src/Rules/PHPUnit/AttributeRequiresPhpVersionRule.php
staabm: alternative approach could be to somehow analyse `#[RequiresPhp('>=8.0')]` as if it would have the AST of `PHP_VERSION_ID >= 80000` - which would re-use the already existing PHP_VERSION_ID checking capabilities

ondrejmirtes:
I'd rather use https://github.com/phar-io/version which is also used by PHPUnit I think https://github.com/sebastianbergmann/phpunit/blob/2d6dc600b52cd74141c422a5d0601ca26c185bb9/src/Metadata/Version/Requirement.php

Yeah, comparing against `PHP_VERSION_ID` would take care of all issues. We just need to properly understand the constraints in all cases.

staabm:
> I'd rather use [phar-io/version](https://github.com/phar-io/version?rgh-link-date=2025-12-08T08%3A45%3A02Z) which is also used by PHPUnit I think [sebastianbergmann/phpunit@`2d6dc60`/src/Metadata/Version/Requirement.php](https://github.com/sebastianbergmann/phpunit/blob/2d6dc600b52cd74141c422a5d0601ca26c185bb9/src/Metadata/Version/Requirement.php)

thats only one part of the story. see also https://github.com/sebastianbergmann/phpunit/blob/2d6dc600b52cd74141c422a5d0601ca26c185bb9/src/Metadata/Version/Requirement.php#L41-L48 which adds a fallback path on top which is implemented in phpunit itself.

reworked it to match that

ondrejmirtes:
The related PR was merged so you can resume your work on this. Thank you.

staabm:
> Yeah, comparing against `PHP_VERSION_ID` would take care of all issues. We just need to properly understand the constraints in all cases.

I got an idea for this one. I will push it tomorrow into this PR

staabm:
@VincentLanglet finally, this is really ready for review :-)

tests/Rules/PHPUnit/AttributeRequiresPhpVersionRuleTest.php
```diff
@@ -78,6 +81,62 @@ public function testRuleOnPHPUnit13(): void
...
 
+	public function testPhpVersionMismatch(): void
+	{
+		$this->phpunitMajorVersion = 12;
+		$this->phpunitMinorVersion = 4;
+		$this->deprecationRulesInstalled = false;
+
+		$this->analyse([__DIR__ . '/data/requires-php-version-mismatch.php'], [
+			[
+				'Version requirement will always evaluate to false.',
```
VincentLanglet, line 94: #[RequiresPhp('<= 8.5')] will always evaluate to false ?

staabm: because the test assumes PHP_VERSION_ID `80500` and the constraint only has 2 digits.
it would not error for `'<= 8.5.0'`.

this is what https://github.com/phpstan/phpstan-phpunit/pull/303 is about

VincentLanglet: That's tricky.

Should we have a test with `8.6` to ensure this one is not reported ?

staabm: done

VincentLanglet (approved):

staabm:
(needs to wait for the PHPStan 2.2.3 release before extension release)

PR #272, opened 2025-12-16
Add support for createMockForIntersectionOfInterfaces

VincentLanglet:


src/Type/PHPUnit/MockForIntersectionDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,79 @@
...
+		$interfaces = $scope->getType($args[0]->value);
+		$constantArrays = $interfaces->getConstantArrays();
+		if (count($constantArrays) !== 1) {
+			return null;
+		}
+
+		$constantArray = $constantArrays[0];
+		if (count($constantArray->getOptionalKeys()) > 0) {
+			return null;
+		}
```
staabm, line 66: this condition seems untested

staabm (commented):
I would expect a return type extension in addition to be tested in a type inference test with assertType()

tests/Rules/PHPUnit/data/mock-method-call.php
```diff
@@ -41,6 +41,22 @@ public function testMockObject(\PHPUnit\Framework\MockObject\MockObject $mock)
 		$mock->method('doFoo');
 	}
 
+	public function testMockForIntersection()
+	{
+		$mock = $this->createMockForIntersectionOfInterfaces([FooInterface::class, BarInterface::class]);
+		$mock->method('fooMethod');
+		$mock->method('barMethod');
```
staabm, line 48: wouldn't a intersection of 2 interfaces only allow calls to methods, which both interfaces contain?
I think what you implemented here is a union behaviour?

VincentLanglet: If you implement both interface, you have both methods

src/Type/PHPUnit/MockForIntersectionDynamicReturnTypeExtension.php
VincentLanglet: Should be ok now

tests/Rules/PHPUnit/MockMethodCallRuleTest.php
```diff
@@ -34,6 +36,17 @@ public function testRule(): void
 			],
 		];
 
+		if (method_exists(TestCase::class, 'createMockForIntersectionOfInterfaces')) { // @phpstan-ignore-line
```
staabm: please ignore by identifier instead of the whole line for all errors

VincentLanglet: done

src/Type/PHPUnit/MockForIntersectionDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,92 @@
...
+{
+
+	public function getClass(): string
+	{
+		return TestCase::class;
+	}
+
+	public function isMethodSupported(MethodReflection $methodReflection): bool
+	{
+		return strtolower($methodReflection->getName()) === 'createmockforintersectionofinterfaces';
```
ondrejmirtes: No need for strtolower here

src/Type/PHPUnit/MockForIntersectionDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	public function isStaticMethodSupported(MethodReflection $methodReflection): bool
+	{
+		return strtolower($methodReflection->getName()) === 'createstubforintersectionofinterfaces';
```
ondrejmirtes: dtto

ondrejmirtes (changes requested):

src/Type/PHPUnit/MockForIntersectionDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,92 @@
...
+			if (!$valueType->isClassString()->yes()) {
+				return null;
+			}
+
+			$values = $valueType->getConstantScalarValues();
+			if (count($values) !== 1) {
+				return null;
+			}
+
+			$result[] = new ObjectType((string) $values[0]);
```
staabm, line 85: could be easy to iterate over all values instead of working only with a single value?

```php
	public function testStubForIntersectionWithUnionArg()
	{
		$method = 'fooMethod';
		if (rand(0,1)) {
			$method = 'barMethod';
		}
		$stub = static::createStubForIntersectionOfInterfaces([FooInterface::class, BarInterface::class]);
		$stub->method($method);
	}
```

VincentLanglet: it's not this use case, but this one
```
public function testStubForIntersectionWithUnionArg()
	{
		$class = FooInterface::class;
		if (rand(0,1)) {
			$class = Foo2Interface::class;
		}
		$stub = static::createStubForIntersectionOfInterfaces([$class, BarInterface::class]);
	}
```

Which I need to understand as `FooInterface&BarInterface | Foo2Interface&BarInterface` and which grows in complexity for any non-single-constant value in the array.

I don't think there is a real usage for this. (and an easy implementation ?)

staabm: I see 👍

src/Type/PHPUnit/MockForIntersectionDynamicReturnTypeExtension.php
VincentLanglet: Done

tests/Rules/PHPUnit/MockMethodCallRuleTest.php
```diff
@@ -34,6 +36,17 @@ public function testRule(): void
 			],
 		];
 
+		if (method_exists(TestCase::class, 'createMockForIntersectionOfInterfaces')) { // @phpstan-ignore-line function.alreadyNarrowedType
```
staabm:
```suggestion
		if (method_exists(TestCase::class, 'createMockForIntersectionOfInterfaces')) { // @phpstan-ignore function.alreadyNarrowedType
```

tests/Type/PHPUnit/MockForIntersectionDynamicReturnTypeExtensionTest.php
```diff
@@ -0,0 +1,42 @@
...
+use PHPUnit\Framework\TestCase;
+use function method_exists;
+
+class MockForIntersectionDynamicReturnTypeExtensionTest extends TypeInferenceTestCase
+{
+
+	/** @return mixed[] */
+	public static function dataFileAsserts(): iterable
+	{
+		if (method_exists(TestCase::class, 'createMockForIntersectionOfInterfaces')) { // @phpstan-ignore-line function.alreadyNarrowedType
```
staabm:
```suggestion
		if (method_exists(TestCase::class, 'createMockForIntersectionOfInterfaces')) { // @phpstan-ignore function.alreadyNarrowedType
```

staabm (approved):
few nits, lgtm otherwise

tests/Rules/PHPUnit/MockMethodCallRuleTest.php
VincentLanglet: Kinda curious about this feedback.

Isn't `@phpstan-ignore-line function.alreadyNarrowedType` clearer ?

I understand it as `Ignore function.alreadyNarrowedType in the current line`
As opposite to `phpstan-ignore-next-line`

Are phpstan-ignore-line and phpstan-ignore the same ?

staabm: my understanding is
- `@phpstan-ignore-line` always ignores all errors of a single line (does not support a by-identifier filter)
- `@phpstan-ignore-next-line` always ignores all errors of a single line (does not support a by-identifier filter)
- `@phpstan-ignore` requires at least one identifier and ignores only errors of such identifier (on the same line)

ondrejmirtes: Exactly. I suspect people do this mistake often but I don't know how to prevent it.

VincentLanglet: I didn't know that, and made the mistake a lot then
https://github.com/search?q=repo%3AVincentLanglet%2FTwig-CS-Fixer+phpstan-ignore-next-line&type=code

Especially because I don't like adding `@phpstan-ignore` on the same line and prefer to ignore before the line (so I use ignore-next-line a lot).

staabm: as far as I remember there is a open feature request (cannot find it right now), which goal is to disallow `@phpstan-ignore-line` and `@phpstan-ignore-next-line` and allow only ignoring errors by identifier (via opt-in config)

staabm: alternatively we could look at `@phpstan-ignore-line` and `@phpstan-ignore-next-line` like we do for `@phpstan-ignore` - but report a error if we find a valid issue-identifier after it

staabm (approved):

VincentLanglet:
This PR need the approval of @ondrejmirtes since he requested changes

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #274, opened 2025-12-23
Implement DynamicCallToAssertionIgnoreExtension

staabm:
as discussed in https://github.com/phpstan/phpstan-src/pull/4654

---

we use a `IgnoreErrorExtension` to prevent slow regex re-validation of ignore patterns like

```
	ignoreErrors:
		- '#^Dynamic call to static method PHPUnit\\Framework\\\S+\(\)\.$#'
```


ondrejmirtes:
Thank you!

PR #277, opened 2026-02-14
DataProviderDataRule: Optimize hot path

staabm:
refs https://github.com/phpstan/phpstan/issues/14072#issuecomment-3901259495

reduces the cost when analyzing non-test methods return/yield/yield-from which is a very very common case.

before PR

grafik

after PR

grafik


ondrejmirtes:
Thank you!

PR #287, opened 2026-03-12
Fix build

staabm:
remove dead code, which errors since https://github.com/phpstan/phpstan-src/pull/5209

before this PR

```
------ ----------------------------------------------------------------------- 
  Line   src/Rules/PHPUnit/AnnotationHelper.php                                 
 ------ ----------------------------------------------------------------------- 
  50     Call to function array_key_exists() with 'annotation' and array{0:     
         non-falsy-string, annotation: non-falsy-string, 1: non-falsy-string,   
         property: non-empty-string, 2: non-empty-string, whitespace: string,   
         3: string, value: string, ...} will always evaluate to true.           
         🪪  function.alreadyNarrowedType                                       
  50     Call to function array_key_exists() with 'whitespace' and array{0:     
         non-falsy-string, annotation: non-falsy-string, 1: non-falsy-string,   
         property: non-empty-string, 2: non-empty-string, whitespace: string,   
         3: string, value: string, ...} will always evaluate to true.           
         🪪  function.alreadyNarrowedType                                       
  50     Strict comparison using === between true and false will always         
         evaluate to false.                                                     
         🪪  identical.alwaysFalse                                              
         💡  Because the type is coming from a PHPDoc, you can turn off this    
         check by setting treatPhpDocTypesAsCertain: false in your phpstan.neo  
         n.                                                                     
  50     Strict comparison using === between true and false will always         
         evaluate to false.                                                     
         🪪  identical.alwaysFalse                                              
         💡  Because the type is coming from a PHPDoc, you can turn off this    
         check by setting treatPhpDocTypesAsCertain: false in your phpstan.neo  
         n.                                                                     
 ------ ----------------------------------------------------------------------- 
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
 	"keywords": ["static analysis"],
 	"require": {
 		"php": "^7.4 || ^8.0",
-		"phpstan/phpstan": "^2.1.41"
+		"phpstan/phpstan": "^2.1.32"
 	},
 	"conflict": {
 		"phpunit/phpunit": "<7.0"
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
 	"keywords": ["static analysis"],
 	"require": {
 		"php": "^7.4 || ^8.0",
-		"phpstan/phpstan": "^2.1.32"
+		"phpstan/phpstan": "^2.1.41"
 	},
 	"conflict": {
 		"phpunit/phpunit": "<7.0"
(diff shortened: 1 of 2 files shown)


VincentLanglet (approved):

ondrejmirtes:
Should I release it alongside 2.1.41 or not necessary?

ondrejmirtes:
Thank you.

staabm:
> Should I release it alongside 2.1.41 or not necessary?

I think its not necessary

PR #305, opened 2026-06-20
Implement ClassAttributeRequiresPhpVersionRule

staabm:
requires https://github.com/phpstan/phpstan-phpunit/pull/304

### DIFF
--- a/rules.neon
+++ b/rules.neon
@@ -16,6 +16,9 @@ conditionalTags:
 	PHPStan\Rules\PHPUnit\DataProviderDataRule:
 		phpstan.rules.rule: %featureToggles.bleedingEdge%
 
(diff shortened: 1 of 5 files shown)


rules.neon
```diff
@@ -16,6 +16,9 @@ conditionalTags:
 	PHPStan\Rules\PHPUnit\DataProviderDataRule:
 		phpstan.rules.rule: %featureToggles.bleedingEdge%
 
+	PHPStan\Rules\PHPUnit\ClassAttributeRequiresPhpVersionRule:
+		phpstan.rules.rule: %featureToggles.bleedingEdge%
```
staabm, line 20: technically its a new rule, but in the end we just add wiring for a pre-existing service (and its bleeding edge only).

therefore I feel confident merging, without a review from ondrej

VincentLanglet (approved):

PR #313, opened 2026-07-04
Fix `#[RequiresPhpunit]` version requirement being evaluated against PHP version

kubawerlos:
Currently, the `#[RequiresPhpunit]` requirement is evaluated against the analysed PHP version, so constraints like `^11.0.0` are reported as "always evaluate to false" even with PHPUnit 11.5 installed.


staabm:
thanks for cleaning up after me. :)

PR #315, opened 2026-07-06
Fix PHPVersion based checks

staabm:
re-use the logic [extracted into ConfiguredPhpVersionRangeHelper](https://github.com/phpstan/phpstan-src/pull/6014) to make sure expressions 
like `PHP_VERSION_ID < 80200` are handled in the same way as `#[RequiresPhp('<8.2.0')]`

in https://github.com/phpstan/phpstan-src/pull/5972#issuecomment-4885668733 a improvement for the error messaeg is requested.. this will ship in a separate PR

### DIFF
--- a/src/Rules/PHPUnit/AttributeVersionRequirementHelper.php
+++ b/src/Rules/PHPUnit/AttributeVersionRequirementHelper.php
@@ -7,17 +7,14 @@ use PharIo\Version\Version;
 use PharIo\Version\VersionConstraintParser;
 use PHPStan\Analyser\Scope;
 use PHPStan\BetterReflection\Reflection\ReflectionAttribute;
+use PHPStan\Php\ConfiguredPhpVersionRangeHelper;
 use PHPStan\Php\PhpMinorVersionIterator;
-use PHPStan\Php\PhpVersion;
 use PHPStan\Rules\IdentifierRuleError;
 use PHPStan\Rules\RuleErrorBuilder;
-use PHPStan\Type\Constant\ConstantIntegerType;
-use PHPStan\Type\IntegerRangeType;
 use function count;
 use function is_numeric;
 use function preg_match;
 use function sprintf;
-use function strpos;
 use function substr_count;
 use function version_compare;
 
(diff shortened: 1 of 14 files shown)


src/Rules/PHPUnit/AttributeVersionRequirementHelper.php
```diff
@@ -168,29 +161,23 @@ public function checkVersionRequirement(array $attributes, Scope $scope): array
...
-		$scopePhpVersion = $scope->getPhpVersion()->getType();
-		if ($scopePhpVersion instanceof ConstantIntegerType) {
-			$v = new PhpVersion($scopePhpVersion->getValue());
-			return [new Version($v->getVersionString())];
-		} elseif ($scopePhpVersion instanceof IntegerRangeType) {
-			if ($scopePhpVersion->getMin() === null || $scopePhpVersion->getMax() === null) {
-				return [];
-			}
-
+		// @phpstan-ignore phpstanApi.method
```
staabm, line 166: I was not sure, whether we would want to have a `@api` on `ConfiguredPhpVersionRangeHelper`.

thats why I used the ignores for now

tests/Rules/PHPUnit/AttributeRequiresPhpVersionRule.neon
```diff
@@ -1,2 +1,4 @@
 parameters:
-	phpVersion: 80500
+	phpVersion:
+		min: 80500
+		max: 80599
```
staabm, line 4: adjusted to replicate the behaviour the tests expected before this PR

tests/Rules/PHPUnit/AttributeRequiresSinglePhpVersionRule.neon
```diff
@@ -0,0 +1,2 @@
+parameters:
+	phpVersion: 80210
```
staabm, line 2: this test is asserting new behaviour.

before this PR we errored on such single-version configs.
[after this PR we skip them](https://github.com/phpstan/phpstan-src/blob/c1f6ad028567128f9b095ccfe1ffc8fceba96fd7/src/Php/ConfiguredPhpVersionRangeHelper.php#L35).

staabm:
for testing purposes, this PR here is [attached to a phpstan-src PR](https://github.com/phpstan/phpstan-src/pull/5972) to see real world impact on the core

VincentLanglet (approved):

PR #316, opened 2026-07-06
AttributeVersionRequirementHelper: Improve error messages

staabm:
refs https://github.com/phpstan/phpstan-src/pull/5972#issuecomment-4885668733

the error messages before this PR reflected 1:1 what PHPUnit reports.
to make the results more readable/understandable we add more context to the messages.

### DIFF
--- a/src/Rules/PHPUnit/AttributeVersionRequirementHelper.php
+++ b/src/Rules/PHPUnit/AttributeVersionRequirementHelper.php
@@ -135,7 +135,7 @@ final class AttributeVersionRequirementHelper
 				}
 
 				if (strpos($attr->getName(), 'RequiresPhpunit') !== false) {
-					$tip = 'PHPUnit version inferred from composer.json. requirements';
+					$tip = 'PHPUnit version inferred from composer.json requirements.';
 				} else {
 					$tip = 'PHP version for analysis inferred from NEON config phpVersion or composer.json requirements. Invoke PHPStan with -vvv to get more details.';
 				}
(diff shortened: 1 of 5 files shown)


tests/Rules/PHPUnit/AttributeRequiresPhpVersionRangeRuleTest.php
```diff
@@ -17,23 +17,23 @@ public function testPhpVersionMismatch(): void
 	{
 		$this->analyse([__DIR__ . '/data/requires-php-version-mismatch.php'], [
 			[
-				'Version requirement will always evaluate to false.',
+				'Version requirement < 7.0 does not match 8.2.0...8.4.0.',
```
staabm, line 22: I am using a version range notation which github.com is also using. so I think it should be familiar to devs

VincentLanglet (approved):

VincentLanglet (approved):

src/Rules/PHPUnit/AttributeVersionRequirementHelper.php
```diff
@@ -129,9 +130,25 @@ public function checkVersionRequirement(array $attributes, Scope $scope): array
 					}
 				}
 
+				if (count($pharIoVersions) < 2) {
+					throw new ShouldNotHappenException();
+				}
+
+				if (strpos($attr->getName(), 'RequiresPhpunit') !== false) {
+					$tip = 'PHPUnit version inferred from composer.json. requirements';
```
ondrejmirtes: typo, the dot belongs at the end

ondrejmirtes (dismissed):
Otherwise 👍

PR #325, opened 2026-08-21
feat: Discourage assert(Not)Empty if "empty" usage is disallowed

### LINKED ISSUE #270: Disallow `assertEmpty` with strict rule `disallowedEmpty` set
hey there

I am currently trying to activate the `disallowedEmpty` rule of the strict package in our code base (over 1k issues 🎉🤯🙈)

While doing that, I found, that the usage of `assertEmpty` is not affected by this rule. Is this intentionally? Or do you think, it would be an addition that actually makes sense and would make it consistent?

Happy to hear your feedback on this 🙂 

best regards 👋

mitelg:
fixes: https://github.com/phpstan/phpstan-phpunit/issues/270

Hey,

first attempt to fix the linked issue. I used Codex for that and I hope it did the job correct :grin: Tests are looking fine for me. 

Happy to get some feedback :handshake: 

Best regards :wave:


rules.neon
```diff
@@ -12,6 +12,8 @@ rules:
 conditionalTags:
 	PHPStan\Rules\PHPUnit\AssertEqualsIsDiscouragedRule:
 		phpstan.rules.rule: [%strictRulesInstalled%, %featureToggles.bleedingEdge%]
+	PHPStan\Rules\PHPUnit\AssertEmptyIsDiscouragedRule:
+		phpstan.rules.rule: %strictRulesInstalled%
```
mitelg: I would like to bind it to the `%strictRules.disallowedEmpty%` toggle as well, but seems not be possible 🤔

staabm: see line 14 `phpstan.rules.rule: [%strictRulesInstalled%, %featureToggles.bleedingEdge%]` how a switch condition can be written for a AND condition

mitelg: @staabm do you know, how to make this test pass? https://github.com/phpstan/phpstan-phpunit/actions/runs/34488522076/job/102908899134?pr=325 :thinking:

mitelg: okay, I added the strict rules extension to the e2e test. not sure if this is applicable :slightly_smiling_face:

rules.neon
```diff
@@ -12,6 +12,8 @@ rules:
 conditionalTags:
 	PHPStan\Rules\PHPUnit\AssertEqualsIsDiscouragedRule:
 		phpstan.rules.rule: [%strictRulesInstalled%, %featureToggles.bleedingEdge%]
+	PHPStan\Rules\PHPUnit\AssertEmptyIsDiscouragedRule:
+		phpstan.rules.rule: [%strictRulesInstalled%, %strictRules.disallowedEmpty%]
```
VincentLanglet, line 16: Should it be also behind `%featureToggles.bleedingEdge%` ?

rules.neon
```diff
@@ -12,6 +12,8 @@ rules:
 conditionalTags:
 	PHPStan\Rules\PHPUnit\AssertEqualsIsDiscouragedRule:
 		phpstan.rules.rule: [%strictRulesInstalled%, %featureToggles.bleedingEdge%]
```
VincentLanglet, line 14: @staabm You implemented AssertEqualIsDiscouraged with https://github.com/phpstan/phpstan-phpunit/commit/10880dad5190e788634a62ebb211802669382f3d

Looking at how AssertEmptyIsDiscouragedRule is configured ; I wonder if it wouldn't be good to be behind `disallowedLooseComparison` here ?

staabm: I am not sure what you are referring to? `AssertEqualIsDiscouraged` is configured with `[%strictRulesInstalled%, %featureToggles.bleedingEdge%]`?

src/Rules/PHPUnit/AssertEmptyIsDiscouragedRule.php
```diff
@@ -0,0 +1,62 @@
...
+			if (!AssertRuleHelper::isMethodOrStaticCallOnAssert($node, $scope)) {
+				return [];
+			}
+		} elseif ($node instanceof FuncCall) {
+			if (!$node->name instanceof Name || !in_array(strtolower($scope->resolveName($node->name)), ['phpunit\\framework\\assertempty', 'phpunit\\framework\\assertnotempty'], true)) {
+				return [];
+			}
+		} else {
+			return [];
+		}
```
staabm, line 50: function based assertions seem to be not supported by other Assert* rules, so I think we shouldn't start here.

it would need a separate PR doing it in all rules (not sure its worth adding,... noone asked for these yet AFAIR)

rules.neon
staabm: good catch, I think you are right.

rules.neon
VincentLanglet: I mean it should be with
```
[%strictRulesInstalled%, %strictRules.disallowedLooseComparison%, %featureToggles.bleedingEdge%]
```

staabm: I think we cannot use `%strictRules.disallowedLooseComparison%` because it would result into a fatal error when strict rules are not installed.

staabm: `%strictRulesInstalled%` is a flag which is available because it is known in phpstan-src.

https://github.com/phpstan/phpstan-src/blob/b3547b6e094d604560bd0f97420f2ac8fcfa0961/conf/config.neon#L92

we cannot depend on other 1st party config flags which are only available when a extension is installed

VincentLanglet: Then https://github.com/phpstan/phpstan-phpunit/pull/325/changes#diff-0874caf5d665ad61748aa19c613530c9c0db6782cb4a80440c4a8ec1024f79e5R16

`[%strictRulesInstalled%, %strictRules.disallowedEmpty%]` is not valid

mitelg: that's why I added `phpstan/phpstan-strict-rules` to the e2e test. the tests would not work otherwise 😁  

so `strictRules.disallowedEmpty` should be removed again, and `featureToggles.bleedingEdge` should be added?

rules.neon
mitelg: done

rules.neon
mitelg: removed `strictRules.disallowedEmpty`

src/Rules/PHPUnit/AssertEmptyIsDiscouragedRule.php
mitelg: removed it. tbh, it was added by LLM and I thought not further about it 😬

